PR #8, opened 2020-08-12
Get site events from photonranch-api instead of config

timbeccue:
Story: https://www.pivotaltracker.com/story/show/173955139
Demo: http://dev.photonranch.org/site/saf/home

The primary task in this PR is to fetch observatory sky events from the backend service in photonranch-api instead of the site config file. 

I also moved some of the logic into the display component. Previously, the data was obtained by the parent and passed in as a prop, but I think this new structure is cleaner. The time column is now sorted to show utc and the local time for the observatory or the user (based on their computer clock). 

The last few changes (from before this sprint cycle) showcase the moment I realized how to spell 'twilight'.

The following screenshot shows the display that you can also find at the demo link included at the top. Next to 'Site Events' in the middle of the page, click the button that says 'show'.
Screenshot from 2020-08-12 15-01-01

### DIFF
--- a/src/components/CreateProjectForm.vue
+++ b/src/components/CreateProjectForm.vue
@@ -9,8 +9,6 @@
                     <b-input class="project-input" v-model="project_name"></b-input>
                 </b-field>
 
-
-
      <b-tabs type="is-boxed">
 
         <b-tab-item label="Targets">
--- a/src/components/CreateProjectForm.vue
+++ b/src/components/CreateProjectForm.vue
@@ -74,6 +74,7 @@
                             <option value="skyflat"> skyflat </option>
                             <option value="screenflat"> screenflat </option>
                             <option value="bias"> bias </option>
+                            <option value="autofocus"> autofocus </option>
                         </b-select>
                     </b-field>
                     <b-field :label="n==1 ? 'Count' : ''" style="width: 130px;">
@@ -168,8 +169,17 @@
                 </b-field>
                 <b-field 
                     v-if="showAdvancedInputs"
+                    label="Meridian Flip">
+                    <b-checkbox v-model="noflip">No Flip</b-checkbox>
+                </b-field>
+                <b-field 
+                    v-if="showAdvancedInputs"
                     label="Autofocus">
                     <b-checkbox v-model="frequent_autofocus">focus more frequently</b-checkbox>
+                </b-field>
+                <b-field 
+                    v-if="showAdvancedInputs" >
+                    <b-checkbox v-model="near_tycho_star">Use Near Tycho Star</b-checkbox>
                 </b-field>
                 <b-field 
                     v-if="showAdvancedInputs"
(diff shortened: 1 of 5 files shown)


src/components/CreateProjectForm.vue
```diff
@@ -166,11 +167,20 @@
...
+                    v-if="showAdvancedInputs"
+                    label="Meridian Flip">
+                    <b-checkbox v-model="noflip">No Flip</b-checkbox>
+                </b-field>
                 <b-field 
                     v-if="showAdvancedInputs"
                     label="Autofocus">
                     <b-checkbox v-model="frequent_autofocus">focus more frequently</b-checkbox>
                 </b-field>
+                <b-field
```
cmccully, line 180: Is there a way to template out this a bit so that you don't have so many copy/paste sections? I know this is often how frontends work but it was a thought.

src/components/SiteEventsModal.vue
```diff
@@ -42,20 +72,72 @@ export default {
...
-    }
+    },
+    methods: {
+        async getSiteEvents() {
+            let url = `https://api.photonranch.org/api/events?site=${this.sitecode}`
+            let site_events = await axios.get(url)
+            site_events = site_events.data
+
+            // Function to convert from dublin julian days to unix time
+            let djd2unix = t => (t-25567.5)*86400*1000
```
cmccully, line 116: So this is just the python pep8 person in me, but it is nicer aesthetically if your spacing around operators is the same between this line and the next.  I don't know what javascript's convention is but you should follow that.

src/components/SiteEventsModal.vue
```diff
@@ -42,20 +72,72 @@ export default {
...
                     visible: false,
                     sortable: true,
                 },
-            ],
+            ]
+        },
+    },
+    data() {
+        return {
+            //site_events: [{} * 26],
```
mgdaily, line 103: Any reason to keep the commented lines in?

src/components/SiteEventsModal.vue
```diff
@@ -42,20 +72,72 @@ export default {
...
+    },
+    methods: {
+        async getSiteEvents() {
+            let url = `https://api.photonranch.org/api/events?site=${this.sitecode}`
+            let site_events = await axios.get(url)
+            site_events = site_events.data
+
+            // Function to convert from dublin julian days to unix time
+            let djd2unix = t => (t-25567.5)*86400*1000
+            let jd2unix = t => (t - 2440587.5) * 86400 * 1000
```
mgdaily, line 117: Some magic numbers here. Could you make these into named variables? Or a descriptive comment would suffice.

src/components/CreateProjectForm.vue
timbeccue: Good suggestion. In this case, I suspect that reducing the line count comes at the cost of higher complexity. But I will keep that idea in mind.

src/components/SiteEventsModal.vue
timbeccue: Nope, good catch.

PR #16, opened 2021-07-28
Feature/plan targets

kciurleo:
This PR adds a page for planning targets for future observing runs at all of the sites and provides users with a reasonable expectation of what their image should look like.

Note: different version of siderealTime (siderealTimetimesens) in helpers.js changed the output to be in degrees instead of hours; I didn't want to touch the original function, but I think that was causing the original eq2altaz function problems


src/views/PlanTargets.vue
```diff
@@ -0,0 +1,198 @@
...
+
+<style lang="css" scoped>
+  .{targlist {
+    margin-top: 100px;
+  }}
+  .columns {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content:center;
+  }
```
timbeccue, line 168: I think the `columns` class is already defined by Bulma. It's generally preferable to add your own custom class for extra styles rather than overload an existing class.

src/utils/helpers.js
```diff
@@ -77,6 +77,83 @@ var helpers = {
         return [hrz_altitude, hrz_azimuth]
     },
 
+    siderealTimetimesens: function(lon, date) {
```
timbeccue, line 80: It would help to specify what type `date` is. I'm not sure if I should pass in a string with some format, or a moment object, etc. 

Also hope I don't sound like a broken record, but I do not understand the name of this function 😅

https://images.techhive.com/images/idge/imported/article/itw/2013/10/23/programmers_hardest_tasks-600x700-100521914-orig.jpg

src/utils/helpers.js
```diff
@@ -77,6 +77,83 @@ var helpers = {
...
+
+        // Local Sidereal Time:
+        lmst = (100.46 + 0.985647*d + lon + 15*UT);
+        if (lmst<0) {
+            lmst = lmst +360
+        }
+
+        return lmst;
+    },
+    eq2altaztimesens: function (ra, dec, lat, lon, date ) {
```
timbeccue, line 128: Full disclosure: my function/variable names are not the best either (and please let me know if you find any I should improve!)

timbeccue (changes requested):
Looks really nice! I like the simplicity of the interface, and ease of use. Good responsive design, and nice code, especially given you haven't done much with Vue before. Also, it's fun to have someone else working on this instead of just me. 

I added some comments with suggestions before we merge and deploy the changes. Hopefully it shouldn't be too much work. Most are minor things, and there are a few usability things like loading default values in the form. 

One other thing is that the lat/lng fields are editable, and I thought that I would be able to find targets for an arbitrary location by submitting coordinates without specifying a site in the dropdown. But it looks like the site selection is required. I think we should make it clearer to the user what they can do: either allow results for coordinates without a site specified, or imply that the site selection is required by displaying the coordinates as text rather than a form field.

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,198 @@
...
+        setTimeout(function(){ document.getElementById("results").scrollIntoView({behavior: 'smooth'}); }, 30);}
+    },
+  },
+};
+</script>
+
+<style lang="css" scoped>
+  .{targlist {
+    margin-top: 100px;
+  }}
```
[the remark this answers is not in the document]
kciurleo: It does absolutely nothing (as far as I know), just a typo!

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,198 @@
...
+        tzinfo: 'my',
+        offset: '',
+        lcltime: '',
+        lcloffset: '',
+      },
+    };
+  },
+  created: function() {
+    const url = "https://api.photonranch.org/api/all/config"
+    axios.get(url).then(response => {
```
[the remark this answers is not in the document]
kciurleo: At one point, I had left the default site as "None" (which fixes the issue about a required selection when inputting your own latitude/longitude), but that wouldn't be great for "load page, click button, get results" - which do you think would be preferable, that or setting mrc (e.g.) as a default?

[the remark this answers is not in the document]
timbeccue: Hmm, how about this:

If a site is selected from the dropdown, add the `disabled` property to the coordinate fields so the value is displayed but users can't change it.

Either with a radio button toggling 'site vs manual coordinates' or with an extra value in the dropdown (like your 'none' option), users can specify a manual coordinate option which re-enables the coordinate fields for user input. 

I think a default value of MRC for a location would be a safe option, but I'll let you make the final call.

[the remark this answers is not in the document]
kciurleo: Fixed all the other stuff but I haven't been able to find a way to make the datetimepicker accept a default value - the usual way of just setting an initial value doesn't seem to work (either at definition or using Vue.set), and I can't find any suggestions online about how to do so.

It's possible to set an initial value for "dateobs" and make the datetimepicker not required, which will still allow for a user to load page, click button, get results, but it might be unintuitive that the date/time isn't visibly picked - any thoughts?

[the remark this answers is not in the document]
kciurleo: Never mind - think I got it working!

src/utils/helpers.js
```diff
@@ -77,6 +77,84 @@ var helpers = {
...
+        *
+        * Equations courtesy of www.stargazing.net/kepler/altaz.html 
+        * and www.aberdeenastro.org.uk/sidereal_time.htm
+
+        *  LST = 100.46 + 0.985647 * d + lon + 15*UT
+        *
+        *       d    is the days from J2000, including the fraction of a day
+        *       UT   is the universal time in decimal hours
+        *       lon is your longitude in decimal degrees, East positive.
+        *       date is a Date object, e.g. "2021-07-28T07:00:00.000Z"
```
timbeccue, line 92: Maybe say this is a javascript Date object, e.g. `new Date()`. Having an example is good but in this case the example looks like a string rather than a date object.

src/utils/helpers.js
```diff
@@ -77,6 +77,84 @@ var helpers = {
...
+
+        // Local Sidereal Time:
+        lmst = (100.46 + 0.985647*d + lon + 15*UT);
+        if (lmst<0) {
+            lmst = lmst +360
+        }
+
+        return lmst;
+    },
+    eq2altazWithDate: function (ra, dec, lat, lon, date ) {
```
timbeccue, line 129: I know my example you copied doesn't document the function args, but I would copy over the 'date' definition here too just to reduce ambiguity. 

By the way, here's a good resource that includes function documentation comments: https://javascript.info/comments. Specifically, the section including 
```
/**
 * Returns x raised to the n-th power.
 *
 * @param {number} x The number to raise.
 * @param {number} n The power, must be a natural number.
 * @return {number} x raised to the n-th power.
 */
function pow(x, n) {
  ...
}
```

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,201 @@
...
+      targlist: '',
+      user: {
+        lat1: '',
+        lon1: '',
+        dateobs: new Date(Math.round(new Date().getTime() / 1800000) * 1800000),
+        tzinfo: 'my',
+        offset: '',
+        localtime: '',
+        localoffset: '',
+      },
```
timbeccue, line 96: Any reason this is in a parent 'user' object? I think it might be cleaner if all these parameters were nested in the same level under the 'data' return. Then we avoid the potential for the reactivity issue we found earlier. It's a minor thing though; if it works for now that's fine.

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,201 @@
...
+        Vue.set(this.site_info, site, {
+          latitude: response.data[site].latitude,
+          longitude: response.data[site].longitude,
+          name: response.data[site].name,
+          siteoffset: response.data[site].TZ_database_name
+        })
+      }
+      Vue.set(this.user, 'localtime', this.site_info.mrc.siteoffset)
+      Vue.set(this.user, 'lat1', this.site_info.mrc.latitude)
+      Vue.set(this.user, 'lon1', this.site_info.mrc.longitude)
```
timbeccue, line 112: In reference to my previous comment, I guess this could be just 
`this.localtime = this.site_info.mrc.siteoffset`, etc 
if not nested under the 'user' object.

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,201 @@
...
+      console.log(error)
```
timbeccue, line 113: Line 116 does what this is trying to do, so no need for this one (plus, `error` isn't defined in this scope).

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,201 @@
...
+            "image": "/targs/DefaultTargetImages/"+this.easylist[i].name.replace(/ /g, "")+".jpg",
+            "ra": this.easylist[i].ra,
+            "dec": this.easylist[i].dec,
+            "starttime": starttime,
+            "altstart": altstart,
+            "altend": altend
+            })
+        }
+      }
+      console.log(diclist)
```
timbeccue, line 156: Remove this for deployment.

public/targs/introtargets.csv
```diff
@@ -0,0 +1,198 @@
+Group,Object Name,Time of Year,RA,DEC
```
timbeccue, line 1: Is this csv file used now that you have the json file?

src/views/PlanTargets.vue
```diff
@@ -0,0 +1,201 @@
...
+    flex-wrap: wrap;
+    justify-content:center;
+  }
+  .the-form {
+    max-width: 326px;
+    margin-left: 8%;
+    margin-bottom: 4%;
+  }
+  .the-page{
+    margin-top:2%;
```
timbeccue, line 178: use padding instead so the page content isn't pushed below the bottom of the screen.

timbeccue (approved):
Looks good! Just a few minor comments to address. When you finish those, feel free to merge the PR and delete the branch (unless there's more stuff you want to implement now). 

Once the branch is merged, it should deploy to the live website. You should see the changes within a few minutes (ctrl-shift-r if necessary to clear the browser cache).

src/views/PlanTargets.vue
kciurleo: It was mostly for my own debugging purposes - fixed though!

PR #17, opened 2021-08-02
Fix plan-targets

kciurleo:
Changes to the "Find Targets" page after Slack discussion:
- Clock displayed in 24h format when UTC selected
- Messier objects and several other common targets added to list
- Nicknames added for targets
- Minimum altitude changed from 35 to 45 degrees for max airmass of 1.6


src/views/PlanTargets.vue
```diff
@@ -137,9 +147,10 @@ export default {
       for (var i = 0; i < this.easylist.length; ++i) {
         var altstart = helpers.eq2altazWithDate(this.easylist[i].ra, this.easylist[i].dec, this.lat1, this.lon1, starttime)[0]
         var altend = helpers.eq2altazWithDate(this.easylist[i].ra, this.easylist[i].dec, this.lat1, this.lon1, endtime)[0]
-        if (altstart>35 && altend>35) {
+        if (altstart>45 && altend>45) {
```
timbeccue, line 150: If the decision was made based off of airmass, then maybe include a comment saying we picked 45 degrees to satisfy 1.6 airmass or something like that.

timbeccue (commented):
Sorry this took me so long to get to! Looks mostly good, I left a few code comments to address.

Also, I think I have a few more questions about the timezone selection options.

1. What's the difference between my timezone and local timezone? If local timezone means the observatory selected, maybe the wording could make that clearer. 

2.  Switching between local timezone and my timezone doesn't appear to change the time for me, even when a site in a different timezone is selected. 

3. This is perhaps a design decision, but do you think it might make more sense if toggling between my timezone/UTC kept the same moment in time? This would mean toggling between different timezones would keep the same suggested targets. Currently it seems to be that toggling between my timezone and UTC keeps the same hour value (5pm and 17:00 UTC), but these are actually different moments in time. The alternative might be to toggle between 8/11/21, 5pm and 8/12/21, 00:00; in other words, keep the same moment in time.

src/components/TargetCard.vue
```diff
@@ -2,9 +2,8 @@
   <div class="target-card">
     <div class="card">
       <img :src="`${target.image}`"/>
-      <div class="target-name">
-        <h2>{{ target.name }}</h2>
-      </div>
+        <h2 class="target-name">{{ target.name }}</h2>
+        <h4 class="target-nickname">{{ target.nickname }}</h4>
```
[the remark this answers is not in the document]
kciurleo: Noted - I tried it a while ago and it looked weird while spanning the whole element, but I figured out my mistake

kciurleo:
Fixed the following:
1. Reworded tz info to be clearer
2. Toggling between tz keeps same moment in time
3. Added "Observatory time" option for custom lat/long, with user inputting the observatory UTC offset as per Wayne's suggestion
Let me know if there's anything else to clean up - I keep getting weird errors about something in the created hook despite not changing anything there, but I can't figure out what's causing them!

src/views/PlanTargets.vue
```diff
@@ -87,13 +97,34 @@ export default {
       targlist: '',
       lat1: '',
       lon1: '',
-      dateobs: new Date(Math.round(new Date().getTime() / 1800000) * 1800000),
+      customobservatoryoffset: '',
```
timbeccue, line 100: The browser console is warning about this value not matching the expected type (should be a number instead of a string). If you change this initial value to 0, it goes away.

src/views/PlanTargets.vue
```diff
@@ -16,12 +16,19 @@
...
           <p class="control is-expanded">
             <b-field label="Longitude">
-              <b-input type="text" id="lon1" v-model="lon1" required :disabled="localtime!='X'"/>
+              <b-input type="text" id="lon1" v-model="lon1" required :disabled="observatorytime!=='X'"/>
+            </b-field>
+          </p>
+        </div>
+        <div v-if="observatorytime =='X'" class="field">
+          <p class="control is-expanded">
+            <b-field label="Observatory UTC Offset (in min)">
```
timbeccue, line 30: I would make this more explicit "...(in minutes)". Otherwise, maybe it is the minimum offset of multiple DST possibilities? Plus, we're not hurting for space so the layout isn't affected.

By the way, any reason you chose to use minutes instead of hours? I've usually seen timezone offsets expressed in hours (ie. we are -7 in California). Could use decimal hours to handle the weird timezone corner cases. But this is not a critical decision.

timbeccue (approved):
Looks great! I left a couple comments but I'd say they are all optional. Feel free to make whatever remaining changes you want and then merge it into dev.

PR #23, opened 2022-01-12
Merge changes from dev

timbeccue:
Including:
Status bar footer automatically collapses when mousing away (request from Wayne)
Show telescope pointing on sky chart
Ignore null status payloads causing bug
Add download button for last 24 hours of images at a site
Add userway accessibility widget
Login with popup to eliminate homepage redirect on login


src/components/ImageDisplay/ImageNavigationToolbar.vue
```diff
@@ -43,20 +85,55 @@
...
   name: "ImageToolbar",
   data() {
     return {
+      isDownloadModalActive: false,
+      zip_download_waiting: false,
     }
   },
 
   methods: {
+    download_fits_previous_24hrs() {
```
darrenhunt2, line 77: The download last 24 hours button is working for me. Are you planning to include a batch download button for images older than 24 hours under the data tab?

darrenhunt2 (commented):
Status bar changes, UserWay widget, download button all work for me as expected. The login with popup is a useful change. Could a user's ad blocker potentially block the login window, or is that not something that could happen with a login popup like this?
The different crosshairs/pointing on the sky chart can be a little difficult to see, but I'm not sure what colors could be better.

src/components/ImageDisplay/ImageNavigationToolbar.vue
darrenhunt2: The downloaded folder is called uploads.zip. This is a very minor thing, but maybe a descriptive name in the future would be better.

timbeccue: I'm planning on making an interface where users can select a start/end time window, and also specify whether they want only small fits, large fits, or the best available. The server supports the additional queries already so I just need to make the UI. But I wanted to throw this up so there's something in the meantime.

timbeccue: And good catch on the download filename. I made a quick adjustment to `<sitename>-<download-request-timestamp>.zip` and open to other ideas if you have them. This is easy to change whenever.

PR #28, opened 2022-03-11
Refactor/plan targets

kciurleo:
This pull request takes the easy target finding component and moves it to the site pages for each of the observatories. Users should be able to flip back and forth between target cards and the sky map view, and change the date and location of an observation to plan ahead. Users can also click on each card to automatically find that target on the sky map and Aladin view.

One thing in particular I changed recently was the form reactivity; it no longer requires a submit button and changes the target results as the different inputs change. If this is less intuitive or more visually busy, I can change it back.


darrenhunt2 (commented):
Looks good and is in a much better spot! The easy targets list updating as settings are changed makes sense to me, I think it's good as you have it now without a submit button. 
My only thoughts right now are to remove the old find targets page at the top and to make the toggle button visible without having to uncheck the LIVE targets checkbox first.

PR #30, opened 2022-03-12
Feature/easy targets

kciurleo:
This pull request takes the easy target finding component and moves it to the site pages for each of the observatories. Users should be able to flip back and forth between target cards and the sky map view, and change the date and location of an observation to plan ahead. Users can also click on each card to automatically find that target on the sky map and Aladin view.

One thing in particular I changed recently was the form reactivity; it no longer requires a submit button and changes the target results as the different inputs change. If this is less intuitive or more visually busy, I can change it back.

Take two on another branch - didn't have time to make the changes Darren suggested, but they are noted!


src/components/sitepages/SiteTargets.vue
```diff
@@ -56,13 +64,16 @@
                 <div class="sidebar-tabs">
                     <div 
                         :class="{'active': activeSidebarTab=='chart settings'}" 
-                        @click="activeSidebarTab='chart settings'"
+                        @click="activeSidebarTab='chart settings'; show_results=false"
```
timbeccue, line 67: Huh I've never thought to update variables here before, nice

src/components/sitepages/SiteTargets.vue
```diff
@@ -249,7 +346,25 @@ export default {
...
+            show_toggle: '',
+
+            // Directly copied from PlanTargets.vue
+            site_info: {},
+            easylist: list,
+            target: {},
+            targlist: '',
+            selected_target_obs: '',
+            lat1: '',
+            lon1: '',
```
timbeccue, line 361: More descriptive variable names would help. In general, it's better to have code that's easy-to-read rather than easy-to-write.

timbeccue (changes requested):
Really great work so far! A few notes in addition to my code comments:

- I liked the margin separating target photos in the old find targets page. Could you add that here too?
- It would be useful to see which target has been selected. This could be just adding a class that puts a border around a target that was clicked. I can explain in more detail if you want. 
- The layout sometimes gets squished and overlaps with the sidebar.
Screen Shot 2022-03-14 at 11 50 03 AM
- Instead of 'easy targets', how about 'common targets'? I think that is more accurate, as some of these targets are hard to image but they are the ones most commonly observed. 
- I think Darren may have alluded this too, but do we want to remove the site agnostic 'find targets' page now?
- If I select a target, the sidebar automatically changes to the 'telescope controls' tab. I think I'd prefer if it stays on the 'easy targets' tab because I might want to quickly toggle back to the skychart view. Thoughts on this?

src/router.js
```diff
@@ -53,7 +53,7 @@ const router = new VueRouter({
     },
 
     { path: '/about', name: 'about', component: About },
-    { path: '/plantargets', name: 'plan targets', component: PlanTargets},
+    // { path: '/plantargets', name: 'plan targets', component: PlanTargets},
```
timbeccue, line 56: Easy to recreate this if we ever need, so let's just delete it for now.

timbeccue (changes requested):
Everything looks good except for one thing that's probably a quick fix. The sizing of the images are still weird for me. Tried in chrome and firefox with multiple zoom settings which makes me wonder if it's something to do with your setup. Have you tried on different browsers? 
Screen Shot 2022-03-25 at 5 01 46 PM

src/components/sitepages/SiteTargets.vue
```diff
@@ -459,9 +682,31 @@ export default {
...
+            }
+        },
+        dateobsutc() {
+            return moment(this.dateobsreal).add(this.offset, 'm').toDate()
+            //Start time of observation date in "UTC" (ignore timezone info in moment obj)
+        },
+        dateobsobs() {
+            return moment(this.dateobsreal).add(this.offset+this.observatoryoffset, 'm').toDate()
+            //Start time of observation date in "observatory time" (ignore timezone info in moment obj)
+        },
```
timbeccue, line 703: these should be snake case too. they are used the same as variables defined in data().

src/components/TargetCard.vue
```diff
@@ -1,59 +1,56 @@
...
+   }
+   .card-text {
+     background-color: black;
+     color: white;
+   }
+   .selected {
+     border: 3px solid white;
+   }
+ </style>
+
```
timbeccue, line 56: Add a newline at the end of the file

timbeccue (approved):
Nice work! Feel free to merge after the minor variable name changes.

PR #31, opened 2022-03-30
Feature/misc tasks

timbeccue:
This is a bunch of small tasks mostly pulled from trello. 

The biggest change is an update to the way the config files are loaded from the server. There used to be a bug where directly loading a site page (e.g. /site/mrc/observe) threw errors and showed a blank page until reloading because the config did not arrive before components tried to use it to render. Now, the config is loaded once in the top level App.vue, and rendering the router component waits until the config is loaded.

Additional changes:
- The form to create calendar events uses the username as a default event name rather than an empty default
- Simulation/test sites are removed from the list of cards on the homepage (but not the map)
- Site cards on the homepage are sorted by longitude
- Removed an error for the phase status that happens when the server has nothing to display yet
- Added back SiteReservationStatus.vue that was accidentally removed earlier during a cleaning session

### DIFF
--- a/src/components/status/PhaseStatusBar.vue
+++ b/src/components/status/PhaseStatusBar.vue
@@ -3,7 +3,7 @@
     <div class="container phase-status-wrapper">
         <div class="phase-status-title">
             <span>Next Events: </span>
-            <span class="timestamp">{{phase_status_sorted[0].timestamp | readable_date}}</span>
+            <span class="timestamp">{{phase_status_sorted[0]?.timestamp | readable_date}}</span>
         </div> 
         <template v-for="(s, index) in phase_status_sorted">
             <span :key="index + 'spacer'" class="spacer" v-if="index"> | </span>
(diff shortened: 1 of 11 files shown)


src/components/status/PhaseStatusBar.vue
```diff
@@ -48,7 +48,8 @@ export default {
     filters: {
         readable_date(timestamp) {
             // don't forget to convert back to ms for js date handling
-            return moment(timestamp * 1000).format("MM/DD HH:mm:ss");
+            let readable = moment(timestamp * 1000).format("MM/DD HH:mm:ss");
+            return readable
         }
```
kciurleo, line 53: Can I ask what the purpose of this change is? It looks like it does the same thing - is this just a matter of convention?

kciurleo (approved):
Looks good to me - nothing jumps out as confusing, and I think these small changes will all benefit the site!

src/components/status/PhaseStatusBar.vue
timbeccue: Oh, I made that change because I was trying out some extra logic in between but ended up removing it and forgot about the original format. You're right, no functional difference.

PR #33, opened 2022-04-07
Add form for users to restart a site

timbeccue:
This PR addresses the trello request here: https://trello.com/c/hkH7A6JZ
Contents pasted below: 

> I think a good place for this is in the expanded tatus in the lower right-hand corner. The name can be Restart Site. Not Reset or Reboot.
Upon pushing it a dialog should appear: "Please say why you want to do this;" And then a text box. If any character is entered in the box then a Proceed button is enabled and the user can trigger the Re-start, otherwise
the dialog and request are dismissed. I will then enter the time, user and reason in a log.

A few small general syntax/naming changes are included as well.

### DIFF
--- a/src/components/DownloadInterface.vue
+++ b/src/components/DownloadInterface.vue
@@ -41,10 +41,7 @@ export default {
     DatetimeWithTimezonePicker,
   },
   props: {
-    site: {
-      type: String,
-      required: true,
-    },
+    site: String
   },
   data() {
     return {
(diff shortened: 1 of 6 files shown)


darrenhunt2 (commented):
No specific feedback from me, looks good. Requiring and logging a reason for restarting is a good way to handle this, at least for now. The minor naming changes made all read clear to me.

kciurleo:
Agree with what Darren said, and it looks like a logical placement for the reset button too!

kciurleo (approved):

darrenhunt2 (approved):

PR #36, opened 2022-04-27
Feature/misc improvements

timbeccue:
A few miscellaneous changes here:
- fixed page size bug that caused the observatories menu to render in a small vertically scrolling slot (thanks for finding this Darren)
- fixed the x-positioning of the observatories menu given the additional menu items to the left
- built a form component for searching sky objects by name. Added this to the telescopes command tab and targets page. 
- added some logic for clearing the object name if coordinates change by more than a few degrees. This is so if I search for m42 and then adjust the pointing slightly, I don't lose the m42 object name in the command. But if I search for m42 and then go to a very different place in the sky, m42 won't accidentally get sent with commands.
- fix a bug where clearing all the drawn shapes on the image (lines, etc.) wouldn't clear everything
- remove some console.log statements

### DIFF
--- a/src/components/sitepages/SiteTargets.vue
+++ b/src/components/sitepages/SiteTargets.vue
@@ -59,7 +59,7 @@
             <div class="targets-page-content-wrapper">
 
                 <div class="ptr-aladin-parent-div">
-                    <div id="aladin-lite-div" @click="sendCoordinatesToSkychart"/>
+                    <div id="aladin-lite-div" @click="set_coordinates_from_aladin"/>
                 </div>
 
                 <div class="sidebar-tabs">
@@ -79,12 +79,7 @@
 
                 <div class="sidebar-tab-content">
                     <div v-if="activeSidebarTab=='telescope controls'"> 
-                        <b-field label="Search for objects...">
-                            <b-input v-model="simbad_query" @keyup.enter.native="submit_simbad_query"></b-input>
-                            <p class="control">
-                                <b-button @click="submit_simbad_query"><b-icon icon="magnify" /></b-button>
-                            </p>
-                        </b-field>
+                        <TargetSearchField v-model="mount_object" label="Search for objects..." @results="handle_object_name_search" />
                         <command-tabs-accordion 
                             :controls="['Telescope', 'Camera']" 
                             :initInstrumentOpenView="0"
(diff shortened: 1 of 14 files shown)


src/store/modules/command_params.js
```diff
@@ -86,8 +87,36 @@ const getters = {
...
+        const large_anglular_difference_degrees = 3
+
+        let old_ra = helpers.hour2degree(state.mount_ra) // convert hours to degrees
+        let new_ra = helpers.hour2degree(val)
+        let dec = state.mount_dec // since dec doesn't change here, use it for both positions
+        let angle = helpers.angular_distance(old_ra, dec, new_ra, dec)
+        if (angle !== NaN & angle > large_anglular_difference_degrees) {
+            state.mount_object = ''
+        }
+
```
kciurleo, line 101: Is this going to pose a problem when users search for another object in Aladin or click on another common target? (Meaning, is this going to clear the `mount_object` value after picking a completely new target because `large_angular_difference_degrees` is going to be large in that case?)

src/components/sitepages/SiteTargets.vue
```diff
@@ -59,7 +59,7 @@
             <div class="targets-page-content-wrapper">
 
                 <div class="ptr-aladin-parent-div">
-                    <div id="aladin-lite-div" @click="sendCoordinatesToSkychart"/>
+                    <div id="aladin-lite-div" @click="set_coordinates_from_aladin"/>
```
kciurleo, line 62: This function name change makes it much clearer

kciurleo (commented):
This all looks pretty good to me, aside from that one question I had about searching new targets on Aladin/clicking new common targets - makes a ton of sense to have this target search field be its own component, and the misc code edits make everything cleaner

src/store/modules/command_params.js
timbeccue: Good question. The solution feels like it could be cleaner, but basically if you search for a new target, the search component simply returns the results; it doesn't set them to vuex yet or anything. So the parent component is responsible for handling the results (reasoning being that perhaps this component is useful in places unrelated to telescope command parameters). 

In order to avoid the problem you're wondering about, the parent needs to take the results and update the coordinate parameters before updating the object name. If you do it in reverse order, then the object name gets erased. 

For example, in Telescope.vue, lines 234-240:
```javascript
    handle_coordinate_search_results(search_results) {
      if (!search_results.error) {
        this.mount_ra = search_results.ra.toFixed(4);
        this.mount_dec = search_results.dec.toFixed(4);
        // make sure to change this after the coordinates, since the object name is cleared 
        // after large changes in the coordinate positions. Details in vuex command_params.
        this.mount_object = search_results.searched_name;
        ...
```

kciurleo: Okay, that's kind of what I figured! So just keep in mind at this point that updating `mount_object` in the proper order should avoid that

PR #37, opened 2022-05-06
Redirect back to previous page after logout

timbeccue:
This PR introduces a change to the logout flow: instead of always redirecting back to the home page, users are redirected back to the page they were on. 

The main reason for this change (besides perhaps being more intuitive to users) is that we do not want users to be able to navigate to other pages from the control room. Before this change, users who log out in the control room browser cannot get back to the control room page to log in again, because the browser is set up in kiosk mode (disabling browser navigation). 

Auth0, the service we use for user authentication, does not support arbitrary redirects post-logout. As a workaround, the logout routine first saves the route in localStorage. Then the user is logged out via Auth0, redirecting to the /logout endpoint. Finally, this endpoint gets the saved route out of localStorage and redirects back there.

### DIFF
--- a/src/mixins/user_mixin.js
+++ b/src/mixins/user_mixin.js
@@ -48,11 +48,11 @@ export const user_mixin = {
     },
 
     logout() {
-      // make sure the logout happens before redirect.
-      // otherwise, the redirect check for authentication might log the user back in.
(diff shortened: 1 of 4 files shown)


kciurleo (approved):
That's a neat workaround with the change to the router! Looks good to me

PR #38, opened 2022-05-06
Feature/ra and poles

kciurleo:
This pull request makes small changes to the targets page (and ra/dec display components):

- RA display defaults to decimal degrees
- Site saves the window's RA and dec display preference for next use
- North Galactic Pole, South Galactic Pole, and Galactic Center added to sky chart
- Common name parser started for the Aladin search bar (currently contains only "ngp" for north galactic pole and "sgp" for south galactic pole, but more names can be added as we catch those missed by Aladin)


public/data/common_names.json
```diff
@@ -0,0 +1,2 @@
+{   "ngp": "North Galactic Pole", 
+    "sgp": "South Galactic Pole"}
```
timbeccue, line 2: Super minor, but you should end json files with a blank line.

public/data/common_names.json
```diff
@@ -0,0 +1,2 @@
+{   "ngp": "North Galactic Pole",
```
timbeccue, line 1: I was imagining this data inside the function, but I like it as a separate json file even better.

src/utils/object_helpers.js
```diff
@@ -0,0 +1,18 @@
+/* Useful sky object related functions */
+var object_helpers = {
```
timbeccue, line 2: var works fine, but it's been superceded and therefore probably better replaced by either const or let (const in this case).

src/components/display/DecDisplay.vue
```diff
@@ -40,14 +40,15 @@ export default {
       showCopyIcon: false,
 
       displayFormatOptions: ["decimalDegrees", "sexagesimalWithUnits", "sexagesimalPlain"],
-      displayFormatSelected: 0, // index of the list above
+      displayFormatSelected: window.localStorage.getItem('dec_display_format'), // index of the list above
```
timbeccue, line 43: Does the code handle the case where nothing has been set to localStorage yet?

timbeccue: Same question for RA too.

timbeccue (changes requested):
Nice overall! Just a few comments.

src/components/display/DecDisplay.vue
kciurleo: I edited the displayFormat() function down below to set the display format to the first option if this returns null - it seemed to work when I opened it in a new browser!

public/data/common_names.json
```diff
@@ -0,0 +1,3 @@
+{   "ngp": "North Galactic Pole", 
+    "sgp": "South Galactic Pole"}
+
```
timbeccue, line 3: Ok I hate to bring it up again because it is such a minor issue, but just fyi, we're looking for a newline character specifically... I think you may have a space in that line hence the red warning symbol appearing here again. Same in galactic_points.json :)

src/components/display/DecDisplay.vue
timbeccue: Oh I see, great! No need to change, but another alternative would be 
`displayFormatSelected: window.localStorage.getItem('dec_display_format') ?? <default value>`
which uses the nullish coalescing operator (cool name) to set a fallback if the first expression has a nullish value (null or undefined).

This is a new variant of the more familiar || operator, which returns the right side if the left is null, undefined, _or also_ evaluates to a falsy value such as 0, '', or an actual `false`. 

Just another technique to have in your toolbox.

timbeccue (approved):

PR #41, opened 2022-06-09
Feature/night log

timbeccue:
This PR extends/finalizes the initial version of the night log feature. 

The night log is a way for observatory owners (currently: PTR admins) to save a note for subsequent visitors to see. 

Messages can be created and seen in the control room. Messages exist for 48 hours or until another one overwrites it. There is no history, simply the latest message. 

These changes include displaying the time that a message was written, and a mechanism (using localstorage) to show messages on first load and keep them minimized by default after a user sees and closes it. 

The clock mixin is simply a reusable way for components to get the current timestamp, updated each second. It was useful here to update the "5 minutes ago" age display on the message headers but will also be useful in other components that need to be reactive with respect to the current time (status age is an example).


src/components/NightLog.vue
```diff
@@ -79,11 +88,20 @@ export default {
...
                 console.log(response)
+
+                // Save the note we just loaded
                 this.note = response.data
-                if (this.note !== '') {
+                this.note_is_loaded = true
+
+                // Check if the user has already seen this note
+                let note_id = this.note.site + this.note.created_timestamp
+                let last_read_note = window.localStorage.getItem('ptr_nightlog_last_read')
```
kciurleo, line 98: If this night log is intended to be used in the control room, does this get logged to the local storage of the user's browser or the shared control room browser? (If it's the latter, would adding the logged in user's ID to the note_id circumvent that?)

kciurleo (commented):
Looks great, just the one question about local storage. The clock mixin in particular seems like it'll be very useful!

darrenhunt2 (commented):
No specific comments from me, looks good! Agree that the clock mixin will be useful for the future.

timbeccue:
Circling back to this project on Wayne's request. 

The night log is a place observatory owners or PTR admins can leave a note visible to any visitors to the /site/<site>/observe page. 

Notes are simple text messages, with a configurable expiration time, after which the note is removed. 

Notes can also be removed manually (by owners/admins) with a 'delete' button, or by authoring a new note, which overwrites the previous one. 

Users who have not yet seen a note will load the page with the note already open. Opened notes float on top of the UI and can be dragged around on the screen. Closing a note will mark it as read; subsequent page visits (with access to the same browser localstorage) will not automatically open the note. However, as long as a note exists for a site, users can reopen by clicking the "night log" button.

timbeccue:
Admin view with a log message available to view
Screen Shot 2023-02-22 at 6 54 19 PM

Admin view when there is no log message
Screen Shot 2023-02-22 at 6 58 44 PM

Normal user view when there is no log message
Screen Shot 2023-02-22 at 7 01 30 PM

Night log new message editor
Screen Shot 2023-02-22 at 6 54 35 PM

A few more examples:
Screen Shot 2023-02-22 at 6 57 02 PM
Screen Shot 2023-02-22 at 6 56 27 PM
Screen Shot 2023-02-22 at 6 58 26 PM

timbeccue:
Merged so that Wayne can use this tonight, but still open to code review/feedback!

src/store/modules/images.js
```diff
@@ -430,14 +430,19 @@ const actions = {
     for (let channel = 0; channel < 3; channel++) {
       const url = base_url + `/infoimage/${site}/${channel + 1}`
       axios.get(url).then(response => {
+        // Handle case if no info image exists
```
darrenhunt2, line 433: Do we have any plans to actually use info images yet?
I had been wondering if it was just better to remove info images if the feature will never get used.

darrenhunt2 (commented):
After going through, I don't have any changes to request right now. The night log seems to be working as expected after testing it, and it looks great! It will likely be easier to determine any further changes after it gets some use at night.

src/store/modules/images.js
timbeccue: I'm not sure, but this update (along with recent changes in photonranch-api) modifies the http response when there's no info images from a 404 (not found) to a 204 (no content) which removes the errors in the browser console. That should help keep this feature entirely out of the way when it's not in use.

PR #43, opened 2022-06-15
Preserve Subpage/Tab Views

kciurleo:
This pull request preserves the subpage, "Targets" sidebar tab, "Observe" tab, and "Controls Accordion" tab that the user is on when changing the observatory site (and in the case of the tabs, when changing the subpage within an observatory site). It also resubmits the location of the new site in the case of the "Targets" subpage so that the sky chart and common targets will be updated accordingly.

Each of these subpages/tabs are stored as a vuex value, and they all use slightly different methods of updating, as the various subpages/tabs are all different buefy components - hopefully this is enough of a spread that if we add more tabs/subpages in the future and want to repeat the process, this can be a jumping off point!

Note: the accordion page edited was "CommandTabsWide.vue" since that's what's currently used in the "Observe" subpage. "CommandTabsAccordion.vue" was not edited.

### DIFF
--- a/src/components/Tabs.vue
+++ b/src/components/Tabs.vue
@@ -55,15 +55,15 @@
       this.select_tab(this.initial_tab_index)
     },
     methods: {
-      select_tab (i) {
+      select_tab (clicked_tab_index) {
         // provide parent component with selected tab index
-        this.$emit('selected-index', i)
+        this.$emit('selected-index', clicked_tab_index)
 
-        this.selectedIndex = i
+        this.selectedIndex = clicked_tab_index
 
         // loop over all the tabs
         this.tabs.forEach((tab, index) => {
-          tab.isActive = (index === i)
+          tab.isActive = (index === clicked_tab_index)
         })
       },
 


darrenhunt2 (approved):
Looks good! This saves a lot of clicks when switching between sites. No specific changes from me, the pages all behaved as I expected while clicking around to try it.

src/components/sitepages/SiteData.vue
```diff
@@ -199,16 +202,21 @@ export default {
...
 
       region_stats_loading: false,
       image_stats_loading: false,
     }
   },
-  methods: {
 
+  mounted() {
+    //default image tools tab, set to analysis in site_config.js
+    this.activeImageToolsTab = this.selected_image_tools_tab;
```
timbeccue, line 216: Where is `this.selected_image_tools_tab` defined? I don't see any other references to it in the file.

src/components/sitepages/SiteTargets.vue
```diff
@@ -73,7 +73,7 @@
                         class="sidebar-tab-button">telescope controls</div>
                     <div 
                         :class="{'active': activeSidebarTab=='common targets'}" 
-                        @click="activeSidebarTab='common targets'; submitForm()"
+                        @click="activeSidebarTab='common targets'; submitForm(); show_common_targets=true"
```
timbeccue, line 76: I think this would be cleaner if you define a click handler method instead of having multiple statements in the html section

src/components/sitepages/SiteData.vue
```diff
@@ -219,7 +227,11 @@ export default {
   watch: {
     show_user_data_only() {
       this.$store.dispatch('images/load_latest_images')
-    }
+    },
+
+    activeImageToolsTab() {
+      this.$store.commit('site_config/setActiveImageToolsTab', `${this.activeImageToolsTab}`)
```
timbeccue, line 233: Looks like you have  
```
`${...}`
```
to convert the variable to a string. Is that needed/would it already be stored as a string?

src/components/sitepages/SiteTargets.vue
```diff
@@ -644,6 +643,10 @@ export default {
         
     },
     watch: {
+        activeSidebarTab() {
+            this.$store.commit('site_config/setActiveTargetTab', `${this.activeSidebarTab}`);
```
timbeccue, line 647: Similar comment about string conversion, is that needed here?

src/mixins/commands_mixin.js
```diff
@@ -336,6 +340,22 @@ export const commands_mixin = {
...
+      set(value) {this.$store.commit('site_config/setActiveTargetTab', value) }
+    },
+    active_image_tools_tab: {
+      get() { return this.selected_image_tools_tab },
+      set(value) {this.$store.commit('site_config/setActiveImageToolsTab', value) }
+    },
+    active_controls_tab: {
+      get() { return this.selected_controls_tab },
+      set(value) { this.$store.commit('site_config/setActiveControlsTab', value) }
+    },
```
timbeccue, line 358: Since these don't relate to sending commands, better to have them outside of the commands mixin. I would lean towards just adding these to each file that uses them, since it's clearer when trying to figure out where those values are coming from.

src/store/modules/site_config.js
```diff
@@ -15,6 +15,10 @@ const state = {
     did_config_load_yet: false,
 
     selected_site: '',
+    selected_subpage: 'home',
+    selected_target_tab: 'telescope controls', 
+    selected_image_tools_tab: 'analysis',
+    selected_controls_tab: 5,
```
timbeccue, line 21: Since these aren't very related to the site config, I think it would be better to put them in a different module. Perfectly fine if you want to make a new one to keep track of UI state like this. I see why it might seem like site_config is the place to put them, since other user selections (selected_enclosure, etc) are here, but those are used specifically for parsing the config files 🙃

timbeccue (changes requested):
Great work! Seems to function well, most comments are more about code structure and organization than how it works. I think the theme here is to separate logically distinct tasks to minimize the tendency for files to eventually get big and unwieldy. Other than that, seems like you are getting comfortable working with vue and this app specifically, nice job!

src/components/sitepages/SiteData.vue
kciurleo: Definitely not needed - holdover from copying from a use with the subpage, which concatenated strings!

src/components/sitepages/SiteData.vue
kciurleo: Might not be clear enough from the comment above it - another store value defined (currently) in the site_config.js file

src/components/sitepages/SiteData.vue
```diff
@@ -31,12 +31,12 @@
 
       <b-tabs v-model="activeImageToolsTab">
 
-        <b-tab-item label="controls">
+        <b-tab-item label="controls" :value="'controls'">
```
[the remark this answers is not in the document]
kciurleo: Not sure why, but it breaks when I change it to `value="controls"` instead of `:value="controls"` - left it for now but could try to rework

src/components/sitepages/SiteTargets.vue
kciurleo: Cleaned it up as best I can - submitForm is used elsewhere and would be a lot of code to copy into a handler as well, but let me know if you had something different in mind

src/components/Tabs.vue
```diff
@@ -56,6 +56,9 @@
     },
     methods: {
       select_tab (i) {
+        // provide parent component with selected tab index
+        this.$emit('selected-index', i)
+
         this.selectedIndex = i
```
timbeccue, line 62: Very minor, but even for something simple like this I think a more descriptive variable name than `i` would be better.

src/store/modules/user_interface.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * @fileoverview This is the vuex store that maintains the state of various
+ * tabs, pages, etc in the user interface.
+ */
```
timbeccue, line 4: Nice to have this description!

timbeccue (approved):
Nice work with this, especially given the wide variety of tabs you had to figure out. Looks great!

PR #46, opened 2022-06-30
Fixed - Rename dev.js, remove hardcoded API calls

kciurleo:
This pull request renames `dev.js` to `api_endpoints.js` - if a different name is better, it'll be easier to find and replace now. This also changes all hardcoded API calls to reference `api_endpoints.js` instead. I originally added a datastream string to `api_endpoints.js`, but found that this raised errors and have since removed it. (The only references to this string were in `src/datastreamer/index.js` and `PhaseStatusBar.vue`, so unlike the api endpoints, it might be fine to not have as a store value?)

Some notes: UserEventsTable.vue, SiteCalendar.vue, and SiteProjects.vue all made reference to an API endpoint that didn't exist (and then...didn't use that for anything in the code), so those references have been removed entirely.

LogStream.vue and user_status_mixin.js make use of a logs_ws_endpoint that's theoretically defined in api_endpoints.js, but no such store value exists. Should this be changed to a different value/endpoint, or is there something I'm not seeing? (Or is this code just unused?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ npm run serve
 
 ## Configuration
 
-Api endpoint urls used by the frontend are configured in the vuex module in `src/store/modules/dev.js`.
+Api endpoint urls used by the frontend are configured in the vuex module in `src/store/modules/api_endpoints.js`.
 This provides a mechanism for dynamically toggling between different urls.
 
 The backend services defined here and used for the main deployment (and set by default) are:
--- a/src/components/AdminTools/RecentS3UploadsTable.vue
+++ b/src/components/AdminTools/RecentS3UploadsTable.vue
@@ -97,7 +97,7 @@ export default {
     get_recent_uploads() {
       this.table_is_loading = true;
       let url = `${
-        this.$store.state.dev.active_api
+        this.$store.state.api_endpoints.active_api
       }/recentuploads?site=${encodeURIComponent(this.selected_site)}`;
       axios
         .get(url)
--- a/src/components/AnalysisTools/HistogramTool.vue
+++ b/src/components/AnalysisTools/HistogramTool.vue
@@ -54,7 +54,7 @@ export default {
     methods: {
 
         getHistogram(useSubregion=true) {
-        const url = this.$store.state.dev.quickanalysis_endpoint + '/histogram-clipped'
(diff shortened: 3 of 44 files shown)


darrenhunt2 (changes requested):
Looks good and will make our lives so much easier in the future. I didn't catch any typos in your replacements throughout the code. 

The main thing I can think of is to include more of the staged URLs in `api_endpoints.js`. Now that we're adding test, dev, and prod stages for all our endpoints, it'll make things slightly quicker for the future by having those already in the `api_endpoints.js`. It also makes it a little more obvious which URLs are options without having to check the other repositories. (Just be careful because the prod stage is different for different services right now, ex. for the calendar it is "calendar" but for some others it's still "prod".) Maybe order them with the production, dev, and test URLs for each going down the list?

kciurleo:
Great suggestion Darren! Updated with current naming scheme for stage API urls - may need to return to this in the future if that naming scheme changes.

darrenhunt2 (approved):
Looks good!

timbeccue (approved):
Great work! Lots of tedious changes but it's a big improvement to the code. 

Just one small suggestion for making the api_endpoints.js module a little clearer. 

A couple ideas for future improvements:

1. Let the URLs be changed from the UI without modifying the code. For each API:
- Replace the three URLs in the state with a single 'stage' value: either prod, dev, or test. 
- Add a field in the UI that lets you change the stage value. This would happen through a mutation that could validate to make sure the stage is valid.
- The above UI field could work well in the dev tools panel in the site observe tab, or in an admin-only dashboard. 
- All code references to the API's URL should happen through a getter rather than accessing the state directly. 
- The getter would have all three possible URLs, and select the correct one based on the stage value. 

2. Set the default stage with environment variables. This will be good when we end up using www.photonranch for prod and dev.photonranch for dev. 
- Using the setup described above, just define an environment variable for the stage for each api, and read that to set defaults in api_endpoints.js.
- I think you can set production environment variables in the Amplify console. Devs can configure env variables on their machine for local development too. 

Neither of these are urgent or particularly important yet. Well, eventually getting the prod/dev stages separated correctly would be good. Mainly just thinking out loud about this code as I'm reviewing it.

PR #50, opened 2022-08-17
Feature/postage update

kciurleo:
This pull request adds a >> button to the postage stamp view and pulls image data from the most recent day instead of the last 25 images. Changes are in the frontend instead of the api endpoints.

It's currently slow - I'm adding this pull request for ease of sharing changes with Wayne, and will look into why it's slow and make some necessary changes and request a review after the fact.


darrenhunt2 (approved):
Looks and works well for me! Thanks for moving the toggle for "My Data" to a much better location in the Data tab, it's significantly easier to find now.
We can always return to improving the speed of this in the future if you want - it may be quicker to do queries once we finish migrating to the science archive.

PR #53, opened 2022-09-07
default binning inclusion in project form

mfitzasp:
Hopefully this code adds in a default option to binning that defaults to default (if that makes sense). This then sends 0,0 through to the site-code


mfitzasp:
Does this do what I think it does? Pretend for a moment this is the second pull request I have made for javascript in my life.

kciurleo (approved):
This makes sense to me - default binning here is managed differently from in the camera so that different sites could use the same project.

darrenhunt2 (changes requested):
Dropdown default option reads '0, 0' instead of default, which is confusing to users, so we should fix that before merging!

darrenhunt2 (approved):

PR #55, opened 2022-09-15
added object name field to Camera expose and piped it to the command …

mfitzasp:
In this I have added an object field to the camera tab in Observe and made sure that the object name filters down to the exposures/command so that they get sent to site (and updated the relevant site-code also and tested with local server)

### DIFF
--- a/src/components/InstrumentControls/Camera.vue
+++ b/src/components/InstrumentControls/Camera.vue
@@ -31,6 +31,16 @@
       <CancelButton :site="sitecode" style="width: 30%; border-left: 1px solid #00be65" class="is-outlined"/>
     </div>
 
+    <b-field horizontal label="Object Name">
+      <b-input placeholder="Enter your object Name here"
+        type="text"
+        min="0"
+        max="64"
+        size="is-small"
+        v-model="object_name">
+      </b-input>
+    </b-field>
+
     <b-field horizontal label="Expose">
         <b-field>
             <b-input name="subject" size="is-small" v-model="camera_exposure" autocomplete="off"></b-input>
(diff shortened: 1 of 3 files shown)


kciurleo (commented):
Michael, I apologize - I think I was just plain wrong when I confirmed `object_name` was what we were looking to change. There's another store value called `mount_object` that is in use in the telescope page (and project creation form) - I'm assuming that's the one that has snuck through the pipeline at some point to show up in fits headers?

We should switch one or the other so that it stays consistent across all places - and I'm happy to help merge those two different variables if you run into trouble.

mfitzasp:
There is actually a reasoning behind the inconsistency. And actually this couldn't be done without having Spyder open with the site code at SRO, the local server and javascript open to really see how each bit interacts and passes through stuff.

There is actually three and they are all slightly different from each other and have reasons for their independant existance. 

**mount_object** is actually where the mount thinks it is pointing typically drawn from the target tab if people have used that tab to move and slew the telescope manually. This is typically a reliable name that can be searched on Aladin.

**object_name** is new and is actually really just for that specific Camera tab. This object_name can be katierulesok46object7 ... it can be any string that takes your fancy. So if you were doing three runs at a particular object, you could name them katie1, katie2, katie3. If object_name is kept empty in camera, it actually defaults to pulling the mount_object variable in the site code. This is *not* where the mount thinks it is pointing (which is mount_object), but a user-defined whatever goes for an object name thing. 

**block['project']['project_targets'][index]['name']** is also a catchy thing. It is a list of names (currently only ever one name by current design) of targets within a project. When the site-code runs a project, it will pull the object name from here.

mfitzasp:
Oh and OBJSPECF is actually a flag that the site-code creates. If object_name is empty and the mount_object is empty, then it is "Unspecified" and gets an objspecf keyword of no. This causes the site code to generate an object name that is made up of the ra and dec coordinates so at least you know where it was pointing. This should actually be pretty rare -- it would really only happen if you turned up and didn't move the scope and didn't put anything in object field in Camera tab... which seems super unlikely now.

mfitzasp:
Ahh so, mount_object goes to CAT_NAME in the fits header, but it will also go to OBJECT in the fits header when an object name is not specified. but mount_object is really CAT_NAME and not OBJECT.... but can be better than nothing when no OBJECT is specified.

kciurleo (approved):

PR #56, opened 2022-09-20
Bugfix/postage

kciurleo:
An attempt at fixing the postage stamp bug - I've left in some commented out console logging for now in case this doesn't fully fix the problem, but this conversion was the only difference I could find between the postage stamp API call (which wasn't showing correct site noon-noon days) and the filtered data API call (which was showing correct site noon-noon days when input).

### DIFF
--- a/src/store/modules/images.js
+++ b/src/store/modules/images.js
@@ -363,6 +363,7 @@ const actions = {
             queryEnd = siteDate
             queryEnd.setHours (12-siteUserDifference, 0, 0, 0)
             queryEnd.setDate(siteDate.getDate() + 1);
+            // below and in the else, comments for testing correct postage query boundaries
             //console.log("later than noon")        
             //console.log(queryStart)
             //console.log(queryEnd)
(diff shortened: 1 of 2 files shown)


timbeccue:
Curious, but also might be good documentation: what is the postage stamp bug? Is it that the image thumbnails were not loading all the images since the previous site noon? What is the intended behavior and what was happening instead?

kciurleo:
Intended behavior was to have the postage stamp view show the whole day instead of a set number of images, with "day" defined as a site's local noon to noon. Actual behavior was that some of the later images in a night were cut off - it seems like potential time zone conversion error.

src/store/modules/images.js
```diff
@@ -361,24 +361,32 @@ const actions = {
 
             // and the end to noon tomorrow
             queryEnd = siteDate
-            queryEnd.setHours (12, 0, 0, 0)
+            queryEnd.setHours (12-siteUserDifference, 0, 0, 0)
             queryEnd.setDate(siteDate.getDate() + 1);
+            //console.log("later than noon")
```
timbeccue, line 366: I think keeping the comments is ok, but maybe add a description above each block of comments saying why they're there (and in what circumstances can they be removed). Otherwise they will probably blend in as regular abandoned comments when we forget this PR.

timbeccue (approved):

PR #60, opened 2022-09-23
Feature/add linting

timbeccue:
This is a collection of many small changes in an effort to get the existing codebase to  minimal standard of consistency.

Our workflow is based around two linters:
- baseline (`.eslintrc.js`): the entire codebase should always pass
- strict (`.eslintrc_strict.js`): an ideal standard that is impractical to apply everywhere due to the number of changes required

All pull requests that merge into dev must pass two tests:
- baseline lint over the whole codebase
- strict lint over all changed files

This allows us to keep a higher standard without requiring a monumental effort to get everything up to spec. It also allows us to more easily introduce changes to our lint configuration without the aforementioned problems. Details for this process are also included in the README. 

Note: the strict linter is currently the same as the baseline linter because this PR changes so many files that it would essentially require all files to pass (negating the benefit the incremental strict linter is designed to provide). 

---

Nearly all changes included in this PR are bringing the repository up to the level of the baseline linter, which it should now pass, so need to closely inspect all the changes. Lots of small improvements were added I was resolving the manual linting errors. But the main files to note are `.eslintrc.js`, `.eslintrc_strict.js` (the two linter configs), `.github/workflows/linter.yml` (the github action that runs the linters), and `package.json` (which includes helper scripts for locally verifying whether the github action linters will pass).


kciurleo (approved):
After looking through the small changes from linting errors, it's really clear that this'll be a huge help (and especially the ability to run locally before a pull request)! Hopefully gone are the days you have to remind me to add a new line at the end of all my files :smile:

For the stricter linter, what are some examples of rules we'd want to include but couldn't apply to the repository at large?

timbeccue:
> For the stricter linter, what are some examples of rules we'd want to include but couldn't apply to the repository at large?

Good question. I think a good starting point is removing some of the rules I've turned off. For example:
```
    "camelcase": "off",
    "eqeqeq": "off",
    "brace-style": "off",
```

Actually camelcase might be tough because in some cases it could affect variables that exist all over the place. Vscode search will probably find them all but could expand the scope of a single file edit. 

But things like enforcing the correct use of === vs == shouldn't be too hard one file at a time. And there are some vue-specific rules in the same section that are probably worth introducing too. 

I haven't put as much thought into this aspect of things, figured we'd update as needed. Very open to suggestions.

package.json
```diff
@@ -5,7 +5,10 @@
   "scripts": {
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
-    "lint": "vue-cli-service lint",
+    "lint:check": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "lint-changes:check": "eslint -c .eslintrc_strict.js $(git diff --name-only --diff-filter=ACMRTUXB origin/dev | grep  -E \"(.js$|.vue$)\")",
+    "lint-changes:fix": "eslint -c .eslintrc_strict.js $(git diff --name-only --diff-filter=ACMRTUXB origin/dev | grep  -E \"(.js$|.vue$)\") --fix",
```
timbeccue, line 11: These scripts are used for locally linting only the files that differ from origin/dev. A nice explanation for the bash can be found here: https://gist.github.com/seeliang/0f0de424d1cdc4541c338f4ee93b7e6a

I'm not perfectly happy that this isn't exactly the same as the process that runs in github actions. The latter uses a pre-existing action from the marketplace because I couldn't get this script to work with the hardcoded comparison branch (maybe it's possible, I didn't spend much time on it). So this won't work as a way of locally checking the strict linter on changes with a branch that's not dev. 

Worst case if it fails, though, is you find some linting issues when the github action runs and then you can fix and resubmit. Or manually run the strict linter on your changes. I think it's an ok compromise for now, and we can automate it later if it ends up causing pain.

PR #63, opened 2022-09-30
changing bz2 to fz in the download function

mfitzasp:
changing from the bz2 files to the fz files.


mfitzasp:
WORLDS SMALLEST REVIEW. Just changing the file extension to point to the new fz files rather than the bz2 files

timbeccue (approved):
Careful with such large changes, might want to break it into multiple PRs next time ;)

PR #68, opened 2022-11-29
Project Form Update

kciurleo:
Quick change to project form as requested by Michael - now includes two checkboxes for "cycle" and "deplete" for each project, both defaulting to true, to help with project management in the site code.

Edited the layout briefly as well, just to put all the checkboxes on the same row (now that we have so many of them!)


src/components/projects/CreateProjectForm.vue
```diff
@@ -189,6 +190,20 @@
...
+
+        <b-field
+          style="margin-left: 2em;"
+          label="Deplete"
+        >
+          <b-checkbox v-model="depleteIsActive" />
+        </b-field>
+
+        <b-field
+          style="margin-left: 2em;"
```
timbeccue, line 202: In general it's preferable to keep literal css out of the html and use classes instead. In this case I see you're just following convention above, and it's a quick addition, so I'm not saying change it here. Just worth thinking about, in the same spirit as our recent linting improvements.

src/components/projects/CreateProjectForm.vue
```diff
@@ -909,7 +924,10 @@ export default {
         lunar_dist_min: false,
         lunar_phase_max: false
       },
-      calendarBaseUrl: this.$store.state.api_endpoints.calendar_api
+      calendarBaseUrl: this.$store.state.api_endpoints.calendar_api,
+
+      cycleIsActive: true,
+      depleteIsActive: true
```
timbeccue, line 930: Maybe you can think of a better place than here, but it would be nice to have some documentation on what these values represent.

timbeccue (approved):
Looks like a straightforward addition. I added a couple of ideas that would improve readability, but they're not critical so I'll leave that judgement call to you. Don't wanna bog down what should be a simple change.

darrenhunt2 (commented):
When unchecking either of the Deplete or Cycle buttons, is the expected behavior that Hours/Degrees will also go from unchecked to checked? I see this happening when testing out. (I just tried and this also happens when unchecking Smart or Long Stack on the current version of the website). I just want to clarify whether that's intended or not. Otherwise looks good to me!

kciurleo:
> 

Definitely not intended! I'll look into that

src/components/projects/CreateProjectForm.vue
```diff
@@ -917,8 +1063,21 @@ export default {
...
+
+    // initialize smart stack and long stack to camera tab values
+    this.project_constraints.longstack = this.longstackIsActive
+    this.project_constraints.smartstack = this.smartstackIsActive
+  },
+  created () {
+    // initialize expiry date to one lunar month from now, and start date to today
+    const today = new Date()
+
+    this.project_constraints.expiry_date.setDate(today.getDate() + 28)
```
timbeccue, line 1075: Try to avoid [magic numbers](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

src/components/projects/CreateProjectForm.vue
```diff
@@ -917,8 +1063,21 @@ export default {
...
+    this.project_constraints.start_date.setDate(today.getDate())
+
+    // converting from user's timezone to UTC
+    this.project_constraints.expiry_date.setMinutes(this.project_constraints.expiry_date.getMinutes() + today.getTimezoneOffset())
+    this.project_constraints.start_date.setMinutes(this.project_constraints.start_date.getMinutes() + today.getTimezoneOffset())
```
timbeccue, line 1080: Best to not use expiry_date and start_date as temporary variables (line 1075-76) used in creating the actual expiry_date and start_date (lines 1079-80). Instead, just define new variables for line 1075-6 and then you can use them for calculating the final value used to set `this.project_constraints.expiry_date`, etc.

darrenhunt2 (approved):
The tooltips are all useful additions and the Start/Expiry date pickers appear to be working nicely! And no more issues with boxes being checked when you modify a different checkbox.

You don't have to worry about this for now because it's a relic from when I worked on the site selection dropdown, but maybe we should reconsider how the list of selected sites is displayed below that dropdown. The default reading "common pool" below the dropdown isn't immediately obvious as to what it means, it almost looks like it's not supposed to be there. Maybe with the new checkboxes below it, it feels less obvious of where it's coming from to me. I'm not sure if anyone else has strong opinions on that. We can come back to this later.
Otherwise looks good right now!

PR #69, opened 2022-12-05
Add drizzle to project form

kciurleo:
Quick request from Michael to include a "drizzle" parameter in the project form.


darrenhunt2 (approved):
Dropdown looks and works as expected. Looks good, no issues from me!

PR #70, opened 2022-12-07
Postage Stamp Bugfix

kciurleo:
This pull request fixes two bugs in the postage stamp latest noon day display: one reported by Michael that users in Australia were a day ahead, and one reported by Wayne that the postage images were cut off around 12 UTC.

I removed the dependence on user time zone and instead relied on site time zone only, and tried to simplify the math as much as possible to avoid errors (and corrected the final conversion to UTC time for the endpoint query).


darrenhunt2 (approved):
Looks good, and your updated comments are all helpful for following. No apparent issues for me while testing it out.

PR #78, opened 2023-02-23
Better night log reset when switching sites

timbeccue:
If the user navigated to a new site using the menu (not editing the url directly), the night log from the old site would persist into the new one. 

This PR offers a cleaner reset and reload of the night log state with a watcher that runs whenever the route changes.


kciurleo (approved):
Looks fine to me - did you mean to leave the `console.log()` in there, though?

darrenhunt2 (approved):

PR #79, opened 2023-03-01
Feature/remove test sites from map

timbeccue:
Two minor changes here. 

First, a request by Wayne to remove the test sites from the world map and site latest-image tiles below. 

Second, a request from Michael to put the target search and slew buttons in their own row directly under the Aladin window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ node_modules
 /dist
 
 # local env files
-.env
 .env.local
 .env.*.local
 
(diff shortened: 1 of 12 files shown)


timbeccue:
Oh, also included is a change to how the site configs are initially loaded which should prevent some duplicate api calls and reduce other minor async-related glitches.

darrenhunt2 (approved):

PR #83, opened 2023-03-20
Change to expect header ra units in hours

timbeccue:
This update adjusts to a change in the way right ascension is retrieved from fits headers. Previously, image RA was sourced from the 'RAHRS' key, which was expressed in decimal hours. Now, the RA is sourced from the 'CRVAL1' key, with units of decimal degrees. 

The header keys don't matter in the frontend, but the units do. The solution here is a simple computed property that transforms the value from degrees to hours, so that it works as normal in the RaDisplay.vue component that expects input in hours. 

Note that this will break the ra display for any images taken before this change was implemented.

### DIFF
--- a/src/components/ImageDisplay/ImageInfoBar.vue
+++ b/src/components/ImageDisplay/ImageInfoBar.vue
@@ -15,7 +15,7 @@
       style="display:flex"
     >
       ra:&nbsp;<ra-display
-        :ra_hours_decimal="ra_hours"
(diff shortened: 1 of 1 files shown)


kciurleo (approved):

kciurleo (approved):
Looks good, especially the comment about this being redundant in the future if we toss old data (which will save future headache about "what on earth does this code do")

darrenhunt2 (approved):

PR #84, opened 2023-03-22
Feature/auto color

kciurleo:
This pr adds the ability to request automatic 3-color images from the frontend. Options for a site should be defined as an array in the filter wheel config settings named `auto_color_options`, with "manual" as one of them (meaning no auto color should be performed). If there are more options than just "manual", an auto color selection will be added to the camera controls. When the selection is changed from its default of "manual," filter selection is disabled.

(Scaling of exposure time/filter choice/etc is handled in the obsy code once the command is received.)

### DIFF
--- a/src/components/InstrumentControls/Camera.vue
+++ b/src/components/InstrumentControls/Camera.vue
@@ -118,6 +118,7 @@
           v-model="filter_wheel_options_selection"
           placeholder="select filter..."
           size="is-small"
+          :disabled="auto_color_options_selection !== 'manual'"
         >
           <option
             v-for="(filter, index) in filter_wheel_options"
@@ -138,9 +139,9 @@
     </b-field>
 
     <b-field
+      v-if="auto_color_options.length > 1"
       horizontal
       label="Auto-Color"
-      v-if="auto_color_options.length > 1"
     >
       <b-field>
         <b-select


src/components/InstrumentControls/Camera.vue
```diff
@@ -137,6 +138,29 @@
       </b-field>
     </b-field>
 
+    <b-field
+      v-if="auto_color_options.length > 1"
+      horizontal
+      label="Auto-Color"
+    >
+      <b-field>
```
timbeccue, line 146: Is this <b-field> tag necessary?

src/components/InstrumentControls/Camera.vue
```diff
@@ -137,6 +138,29 @@
...
+        <b-select
+          v-model="auto_color_options_selection"
+          placeholder="auto-color..."
+          size="is-small"
+        >
+          <option
+            v-for="(color_opt, index) in auto_color_options"
+            :key="index"
+            :value="color_opt"
+            :selected="index === 0"
```
timbeccue, line 156: I don't think this is necessary because the default selection is already defined through the v-model setting.

src/store/modules/site_config.js
```diff
@@ -187,6 +187,13 @@ const getters = {
     return fwo
   },
 
+  // Available automatic 3-color image options
+  auto_color_options: (state, getters) => {
+    const fwo = getters.selected_filter_wheel_config.settings?.auto_color_options
```
timbeccue, line 192: fwo (from the method above) refers to filter wheel options (admittedly not a fantastic name). I think a different name here would help future readability.

src/store/modules/site_config.js
```diff
@@ -187,6 +187,13 @@ const getters = {
...
+    if (fwo == undefined) return [[]]
```
timbeccue, line 193: Generally a good idea to use the strict equality operator `===` to check if something is undefined. The loose equality check `==` will equate `undefined` and `null`, which usually isn't intended behavior. (I know, my mistake in the method above too.)

Also, while the filter wheel options expects a different format (the first array element is an array of keys, so an empty value still needs one element). But the auto_color_options is just a simple array of values. So the empty value should just be a simple `[]` rather than `[[]]`.

timbeccue (changes requested):
Just a few minor changes, but the functionality seems to work fine!

PR #96, opened 2023-09-14
Feature/group related images

capetillo:
### DESCRIPTION

Background:
Due to an increasing number of thumbnail images making the UI difficult to navigate and not as appealing as it could be, @wrosing requested a feature that groups images by their "smart stack" and only displays the most recent image from each stack. This minimizes the number of thumbnails and simplifies the user experience.

Implementation:
We (@mgdaily and myself) have introduced a new object named grouped_images to organize these images efficiently. This object contains two keys:

1. _site_: The value for this key is the currently selected site, sourced from the rootState. This value resets when the user selects a different site.
2. _imageGroups_: This is a nested object where each key is the SMARTSTK label fetched from image headers. The value is an array containing all image headers with the same SMARTSTK. The array gets populated conditionally based on whether a SMARTSTK key already exists. If SMARTSTK is equal to 'no', then the basefile_name found in the headers becomes the key. This is to prevent losing images that are not in a smart stack and avoid grouping them.

Features:

1. Reduced Thumbnail Count: Only the most recent image from each smart stack is displayed in the ThumbnailRow component. Thus, reducing significantly the number of thumbnail images displayed.
2.  Error Handling: We've maintained the functionality where a default error image displays when no images are available or while loading.

Visuals:
**Before grouping images**

Before grouping images

**Afer grouping images**

After grouping images

**Error/Loading state**

Screenshot of what it looks like when there is no image to display/loads


src/components/sitepages/SiteData.vue
```diff
@@ -308,6 +388,7 @@ import CommandButton from '@/components/FormElements/CommandButton'
 import ImageInfoBar from '@/components/ImageDisplay/ImageInfoBar'
 import InfoImageThumb from '@/components/ImageDisplay/InfoImageThumb'
 import ThumbnailRow from '@/components/ImageDisplay/ThumbnailRow'
+import GroupImagesButton from '@/components/ImageDisplay/GroupImagesButton'
```
mgdaily, line 391: Are we using the GroupImagesButton for this feature? I remember it being designed for the "exploding view" that we want to do later. Would it be safe to exclude from this PR?

src/components/sitepages/SiteData.vue
```diff
@@ -25,156 +25,236 @@
           <thumbnail-row
             class="data-thumbs"
             :images="recent_images"
+            :grouped_images="grouped_images"
             :selected_image="current_image.image_id"
             @thumbnailClicked="setActiveImage"
           />
+          <group-images-button
```
mgdaily, line 32: Can we exclude this from *this* PR since we're not using it at the moment?

src/store/modules/images.js
```diff
@@ -259,6 +266,42 @@ const actions = {
      *     taken the active user or not.
      */
 
+  // we want to group images as they load and we group them based on their SMARTSTK value
+  // we dispatch this action in multiple places
+  group_images ({ commit, state, rootState, dispatch }) {
+    const currentSite = rootState.site_config.selected_site
+    let grouping_images = JSON.parse(JSON.stringify(state.grouped_images))
```
mgdaily, line 273: Nitpick, but can we rename `grouping_images` to `grouped_images`? I think it's clear that the variable has local scope and won't be confused with the global state variable.

src/store/modules/images.js
```diff
@@ -259,6 +266,42 @@ const actions = {
...
+    const recent_images = state.recent_images
+    if (grouping_images.site && grouping_images.site !== currentSite) {
+      dispatch('reset_grouped_images')
+      grouping_images = { site: currentSite }
+    } else if (!grouping_images.site) {
+      grouping_images.site = currentSite
+    } else if (grouping_images.site && grouping_images.site == currentSite) {
+      console.log('continuing')
+    }
+    for (let i = 0; i < recent_images.length; i++) {
```
mgdaily, line 283: Can we use a `for...of` loop here? It's generally cleaner I think than using a traditional indexed loop

src/store/modules/images.js
```diff
@@ -423,20 +467,22 @@ const actions = {
 
       commit('setCurrentImage', response[0])
       commit('setRecentImages', response)
+      dispatch('group_images')
+      // commit('setGroupedImages', { grouped_images })
```
mgdaily, line 471: Unnecessary comment?

src/store/modules/images.js
```diff
@@ -423,20 +467,22 @@ const actions = {
...
 
-  load_latest_info_images ({ state, commit, rootState }) {
+  load_latest_info_images ({ state, commit, rootState, dispatch }) {
     const site = rootState.site_config.selected_site
     const base_url = rootState.api_endpoints.active_api
-
     // query each of the three channels
     for (let channel = 0; channel < 3; channel++) {
       const url = base_url + `/infoimage/${site}/${channel + 1}`
+      console.log('this is channel,', channel)
```
mgdaily, line 485: Extra console log

src/store/modules/images.js
```diff
@@ -447,6 +493,7 @@ const actions = {
         // Don't want to yank the focus from the user
         if (state.current_image.s3_directory == 'info-images') {
           commit('setCurrentImage', response.data)
+          // dispatch('group_images')
```
mgdaily, line 496: Can we remove this commented line?

src/store/modules/images.js
```diff
@@ -458,9 +505,11 @@ const actions = {
   async loadCurrentImageFitsHeader ({ state, rootState, commit }) {
     if ('base_filename' in state.current_image) {
       const baseFilename = state.current_image.base_filename
+      console.log('this is basefilename', baseFilename)
```
mgdaily, line 508: Extra console log

src/store/modules/images.js
```diff
@@ -458,9 +505,11 @@ const actions = {
...
 
       // Check if the header is already cached
       if (state.cached_headers[baseFilename]) {
+        console.log('cached headers', state.cached_headers[baseFilename])
```
mgdaily, line 512: Extra console log

src/store/modules/images.js
```diff
@@ -509,8 +558,8 @@ const actions = {
   },
 
   /**
-     * Set the current image to the most recent one in recent_images.
```
mgdaily, line 512: Formatting weirdness

src/store/modules/images.js
```diff
@@ -541,6 +590,11 @@ const actions = {
     commit('setCurrentImage', first_image)
   },
 
+  set_grouped_images ({ commit, state }) {
+    const grouped_image = state.grouped_images
```
mgdaily, line 594: for consistency can we name `grouped_image` `grouped_images` instead?

src/store/modules/images.js
```diff
@@ -259,6 +266,42 @@ const actions = {
...
+    } else if (!grouping_images.site) {
+      grouping_images.site = currentSite
+    } else if (grouping_images.site && grouping_images.site == currentSite) {
+      console.log('continuing')
+    }
+    for (let i = 0; i < recent_images.length; i++) {
+      const img = recent_images[i]
+      const header = img && img.header
+      let SMARTSTK = header && header.SMARTSTK
+      if (SMARTSTK === 'no') {
```
mgdaily, line 287: Maybe add a comment about what we're doing when the SMARTSTK header keyword is 'no'

src/store/modules/images.js
```diff
@@ -259,6 +266,42 @@ const actions = {
      *     taken the active user or not.
      */
 
+  // we want to group images as they load and we group them based on their SMARTSTK value
+  // we dispatch this action in multiple places
+  group_images ({ commit, state, rootState, dispatch }) {
+    const currentSite = rootState.site_config.selected_site
+    let grouping_images = JSON.parse(JSON.stringify(state.grouped_images))
+    const recent_images = state.recent_images
+    if (grouping_images.site && grouping_images.site !== currentSite) {
```
mgdaily, line 275: Can we add a comment about why we're checking the site?

src/components/ImageDisplay/ThumbnailRow.vue
```diff
@@ -1,22 +1,37 @@
...
     >
+      <img
+        v-if="SMARTSTK !== 'site'"
+        :src="thumbnailWithFallback(item)"
+        onerror="this.onerror=null;this.src='https://via.placeholder.com/60/FF0000/FFFFFF?text=jpg'"
+        :title="item[0].baseFilename"
+        :class="{'selected_thumbnail' : item[0].image_id == selected_image}"
+        loading="lazy"
+        class="recent-image"
+        alt="heck"
```
mgdaily, line 19: maybe change the alt text ;)

src/components/ImageDisplay/ThumbnailRow.vue
```diff
@@ -1,22 +1,37 @@
...
-      v-for="(item, index) in images"
-      :key="index"
-      :src="thumbnailWithFallback(item)"
-      onerror="this.onerror=null;this.src='https://via.placeholder.com/60/FF0000/FFFFFF?text=jpg'"
-      :title="item.base_filename"
-      :class="{'selected_thumbnail' : item.image_id == selected_image}"
-      loading="lazy"
-      class="recent-image"
-      @click="setActiveImage(item)"
+  <div
```
mgdaily, line 2: Do we need these extra divs? Can we design our loop a bit like what was there before? I notice you're only showing the parent div if there are grouped images - AFAIK the old code still shows the image carousel with the "no jpg available" even when there aren't images - I think we want to replicate that

src/components/ImageDisplay/ThumbnailRow.vue
```diff
@@ -1,22 +1,37 @@
...
+    v-if="Object.keys(grouped_images).length > 0"
+    class="images"
+  >
+    <div
+      v-for="(item, SMARTSTK) in grouped_images"
+      :key="SMARTSTK"
+      class="image-container"
     >
+      <img
+        v-if="SMARTSTK !== 'site'"
```
mgdaily, line 12: So this is partially my bad, but having to check our SMARTSTK value feels like a code smell - I think it's going to confuse future-me. As it stands, we have a JSON object of:

```
{'site': mrc1,
 1234: [...grouped images...],
 5678: [...grouped images...]
}
```

etc...

It might be cleaner to have a structure like 

```
{'site': mrc1,
 'imageGroups': [ 
    1234: [...grouped images...],
    5678: [...grouped images...],
    ...]
}
```

so that your v-for loop becomes `v-for="item in grouped_images.imageGroups"` and then you don't have to check if the SMARTSTK key is site! I don't think you'll even need the SMARTSTK key anymore

src/components/ImageDisplay/ThumbnailRow.vue
```diff
@@ -25,40 +40,56 @@ export default {
...
+  width: auto;
+}
+
 .recent-image {
   cursor: pointer;
   margin: 0;
   margin-bottom: 5px;
-  flex: 0 0 auto;
-  height: 60px;
+  width: 120px;
```
mgdaily, line 83: Are these CSS changes needed?

src/components/ImageDisplay/GroupImagesButton.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+  <div>
+    <div
+      v-for="(imageList, SMARTSTK) in grouped_images"
+      :key="SMARTSTK"
+    >
+      <!-- Show the last image by default -->
+      <!-- <div v-if="!showAllImages[SMARTSTK]"> -->
```
mgdaily, line 8: Please correct if wrong - but AFAIK we're not using this component for this small feature, so can we remove it from this PR?

mgdaily (changes requested):
Nice work on this! 

I've been quite thorough here and tried to capture everything I see - I think we're on the right track but there's a few changes we can probably make to make this even better!

src/store/modules/images.js
mgdaily, line 485: Can we remove this?

src/components/sitepages/SiteData.vue
capetillo: Deleted!

src/store/modules/images.js
capetillo: Renamed to `grouped_images_local` since it can't be named grouped_images because that already exists

src/store/modules/images.js
capetillo: Great idea, done! Also got rid of else if statement on lines 280-282

src/store/modules/images.js
capetillo: Lovely suggestion, done!

src/store/modules/images.js
capetillo: yep!

src/store/modules/images.js
capetillo: Done!

src/components/ImageDisplay/ThumbnailRow.vue
capetillo: Great idea. Seems a lot more organized. Done!

src/components/ImageDisplay/ThumbnailRow.vue
capetillo: Fixed this

src/components/ImageDisplay/GroupImagesButton.vue
capetillo: we're not using it but I get this error when deleting key "Elements in iteration expect to have 'v-bind:key' directives."

src/components/sitepages/SiteData.vue
```diff
@@ -25,156 +25,236 @@
...
         </div>
 
         <ButtonRowBelowImage class="mt-3" />
       </div>
 
       <!-- Collapsible panels on the right of the image -->
       <div class="image-tools-area">
-
-        <div v-if="userIsAdmin" class="obs-config-box">
+        <div
```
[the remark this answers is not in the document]
capetillo: I get errors if the format is any different. My apologies. I can try to figure this one out with you

src/components/sitepages/SiteData.vue
capetillo: nope, deleted!

src/store/modules/images.js
capetillo: Fixed this so it's easier on the eyes

src/components/ImageDisplay/ThumbnailRow.vue
capetillo: fixed it!

markBowman:
Nicely done!

mgdaily (approved):
yay! nice work!

PR #102, opened 2023-09-28
Remove redundant api calls for fits headers

timbeccue:
Previously, accessing the fits header for an image required an api call to fetch it. Since header data is now included when images are loaded, we don't need to make a separate call each time we want to access a new image's header info.

This change removes the api calls from the fits header modal component, and uses the header data that is already included in the image's metadata object.


capetillo (approved):
I don't think this will affect anything else. Looks good!

PR #104, opened 2023-10-04
Feature/update mosaic fields

capetillo:
### UPDATE: Added two methods to be able to add symbols to the width, height, and angle mosaic fields

**Implementation**
The first method, `getSymbol`, gets the appropriate symbol depending on whether `Mosaic arcmin.` or `Mosaic deg.` is selected under the `Zoom` field. If anything else is selected, then no symbol is added. The symbol is added by dynamically changing the classes of `Width` and `Height`. The symbol is added via CSS by changing `content` to either a ° or a '

The second method, `addSymbol`, adds a static class to the `Angle` field. We can't just add a class to the `Angle` input because if we add the content of ° to the class, then the whole input disappears. For this reason, this method simply returns 'angle-input' and becomes the class of the field. This way we can append the symbol inside the box and expect normal behavior. 

**Notes**
I tried adding a span and other html elements, but that just renders the symbol outside of the box. I believe this is related to buefy but in the meantime, this seems to work.

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -369,6 +369,7 @@
             >
               <b-input
                 v-model="exposures[n-1].width"
+                :class="getSymbol(exposures[n-1].zoom)"
                 size="is-small"
                 :disabled="!exposures[n-1].active"
                 type="number"
@@ -383,6 +384,7 @@
             >
               <b-input
                 v-model="exposures[n-1].height"
+                :class="getSymbol(exposures[n-1].zoom)"
                 size="is-small"
                 :disabled="!exposures[n-1].active"
                 type="number"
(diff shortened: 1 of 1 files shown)


src/components/projects/CreateProjectForm.vue
```diff
@@ -397,13 +399,15 @@
...
                 v-model="exposures[n-1].angle"
+                :class="addSymbol(exposures[n-1].angle)"
                 size="is-small"
                 :disabled="!exposures[n-1].active"
                 type="number"
                 min="-90.0"
                 max="90.0"
                 step="5"
               />
+              <template v-slot:append>°</template>
```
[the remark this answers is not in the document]
capetillo: got rid of this!

src/components/projects/CreateProjectForm.vue
```diff
@@ -397,6 +399,7 @@
             >
               <b-input
                 v-model="exposures[n-1].angle"
+                :class="addSymbol(exposures[n-1].angle)"
```
mgdaily, line 402: that's so weird that setting `:class="angle-input"` makes the input disappear?!

capetillo: it's not the class, it's the content of the class. I'm trying to understand why that's the case but when the class is a method and the content is a symbol, it works 🤔

mgdaily (approved):
Looks good! Just one nitpicky comment. Please teach me your CSS ways 🙏

src/components/projects/CreateProjectForm.vue
mgdaily: hmmmmm

PR #105, opened 2023-10-04
AddSymbol method was not necessary.

capetillo:
There was a typo. My apologies. Added class and content and that does the trick!


capetillo:
got rid of a method. my apologies

PR #109, opened 2023-10-16
Feature/owm report mirrored

LTDakin:
Problem:  Users wanted to see OWM reports on the home tab of site pages. 

Solution: 

- Created a component out of the OWMReport data and copied it over to the home tab. 

- Carried over property of the data only being visible to Admins. 
- In addition improved performance of app by changing the behavior of the OwmReport. Previously it would call the api every time it was mounted. Now, api calls are made in the store so they can be cached. The component will request its data from the store and the api will only be called if an hour has passed since the last report was received. 
- Added Vuex-persistedstate module
- Using persisted state module to store owmReport in localstorage

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -20,9 +20,12 @@ import createPersistedState from 'vuex-persistedstate'
 
 Vue.use(Vuex)
 
-// Todo: research more into other state variables could benefit from a persisted state and seperate into a module
+/**
+ * Todo: research more into other state variables could benefit from a persisted
+ * state and seperate into a module persisting this state to save api calls when page refreshes
+ */
 const dataState = createPersistedState({
-  paths: ['sitestatus.owmReport']
+  paths: ['sitestatus.siteOwmReports']
 })
 
 const store = new Vuex.Store({
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -20,6 +20,7 @@ import createPersistedState from 'vuex-persistedstate'
 
 Vue.use(Vuex)
 
+// Todo: research more into other state variables could benefit from a persisted state and seperate into a module
 const dataState = createPersistedState({
   paths: ['sitestatus.owmReport']
 })
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -16,12 +16,18 @@ import user_interface from './modules/user_interface'
 import uiSync from './modules/uiSync'
 
 import UiSyncPlugin from './plugins/ui_sync'
+import createPersistedState from 'vuex-persistedstate'
 
 Vue.use(Vuex)
 
+const dataState = createPersistedState({
+  paths: ['sitestatus.owmReport']
+})
+
 const store = new Vuex.Store({
   plugins: [
-    UiSyncPlugin
+    UiSyncPlugin,
+    dataState
   ],
   modules: {
     site_config,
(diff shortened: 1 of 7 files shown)


src/store/index.js
```diff
@@ -16,12 +16,19 @@ import user_interface from './modules/user_interface'
 import uiSync from './modules/uiSync'
 
 import UiSyncPlugin from './plugins/ui_sync'
+import createPersistedState from 'vuex-persistedstate'
 
 Vue.use(Vuex)
 
+// Todo: research more into other state variables could benefit from a persisted state and seperate into a module
```
mgdaily, line 23: Perhaps add a bit to this comment saying that this is used to persist the OWM report state across page refreshes - i.e. _why_ we're doing this

src/store/modules/sitestatus/index.js
```diff
@@ -30,7 +30,7 @@ const state = {
 
   weather: {},
   enclosure: {},
-
+  owmReport: [],
```
mgdaily, line 33: This variable name isn't super descriptive - maybe something like siteOwmReports would make more sense, as it's storing multiple owmReports per site.

Something to think about - would it make sense to have this be a JSON object, with each key being a site code? It would make the lookup simpler, so you can just do it by key.

e.g. 

```
{'sro1': {'report': ... ,
          'timestamp': ...},
 'aro': {'report': ... ,
         'timestamp': ...},
}
```

src/store/modules/sitestatus/index.js
```diff
@@ -267,6 +273,14 @@ const mutations = {
     state.obs_settings = status.obs_settings
   },
 
+  new_owmReport (state, newOwmReport) {
```
mgdaily, line 276: the function name here could be more descriptive - I think `storeNewOwmReport` would make more sense

src/store/modules/sitestatus/index.js
```diff
@@ -403,6 +416,26 @@ const actions = {
     }
   },
 
+  getLatestOwmReport ({ commit, rootState, rootGetters, state }) {
+    const wema_name = rootGetters['site_config/wema_name']
+    if (wema_name) {
+      const owmReportObj = state.owmReport.find(owmReport => owmReport.wema_name === wema_name)
+      // request and store a new report if not cached or cached more than 1 hour ago
+      if (owmReportObj == undefined || (Date.now() - owmReportObj.timestamp) > (1000 * 60 * 60)) {
```
mgdaily, line 424: I see what you're doing here, but we generally look at 1000, 60, and 60 here as magic numbers since at a glance they're not super meaningful. It looks like we have the moment library installed in the app, so a cleaner way to do this could be to use that.

https://stackoverflow.com/a/42529483

src/store/modules/sitestatus/index.js
```diff
@@ -94,22 +101,22 @@ const getters = {
...
-      // enclosure is stale
+        // enclosure is stale
       } else if (!enclosure_not_stale && weather_not_stale) {
         return {
           text: 'enclosure not reporting',
           colorClass: 'is-yellow'
         }
-      // weather is stale
-      } else if (!enclosure_not_stale && weather_not_stale) {
+        // weather is stale
```
mgdaily, line 116: These formatting changes seem unrelated to the feature, are they needed for this?

src/store/modules/sitestatus/index.js
```diff
@@ -94,22 +101,22 @@ const getters = {
...
+      } else if (enclosure_not_stale && !weather_not_stale) {
         return {
-          text: 'enclosure not reporting',
+          text: 'weather not reporting',
```
mgdaily, line 119: Why did we change this?

mgdaily (changes requested):
Nice. Still a couple of questions

mgdaily (approved):

PR #117, opened 2023-11-03
Fix/image carousel display

capetillo:
# FIX: Incorrect image display 

## Description

**Background**:
I discovered a bug where the displayed images and image carousel did not correspond to the selected site if the user clicked through sites quickly. 

**Implementation**:
I added `lastSiteRequested` to the state and initialized it to `null`. The state gets mutated to the selected site when the user switches sites. This happens in `load_latest_images` in images.js. The purpose of this is to check if `lastSiteRequested` is the same as `rootState.site_config.selected_site` (i.e. `site` in this function). If it is, then proceed with the api request. If it isn’t, then don’t make the api request. 

## Visuals:
**Before fixing bug, if user clicked through sites quickly, the image to be displayed would be aro1**
Screenshot 2023-11-01 at 10 58 34 AM

**After fixing issue, user can click through sites without erroneous behavior**

https://github.com/LCOGT/ptr_ui/assets/54489472/4469a0c5-55a8-444c-87b8-b699c9ed6a39

### DIFF
--- a/src/store/modules/images.js
+++ b/src/store/modules/images.js
@@ -369,7 +369,7 @@ const actions = {
       method: 'GET',
       url: rootState.api_endpoints.active_api + '/' + site + '/latest_images/1'
     }
-
+    // Doing a try...catch block instead of an await because we get better response time
     try {
       let response = await axios(firstbody)
       response = response.data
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
Looks okay to me and seems like the right idea to block setting the images/data if its not the currently selected site.

PR #121, opened 2023-11-13
Auto loading min zenith distance from site_config data

LTDakin:
Feature request: the automatic loading from config of zenith value if it exists

Implementation: 
This method is called on createProjectForm page mount. It will look and see if there is a default_zenith value for that site and set the project params store equal to it. Since the form is v-modeled from the project params vuex store, this change will appear on the ui box min zenith distance as well

This 5 value was loaded from site_config>mount config>default zenith. If the mount config or zenith default do not exist the value defaults to 0
Screenshot 2023-11-13 at 2 05 16 PM

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -862,6 +862,9 @@ export default {
 
     // set default filter
     this.exposures[targets_index].filter = this.default_filter_selection
+
+    // loading min_zenith_dist fromt config
+    this.load_default_zenith_from_mount()
   },
   watch: {
     // This runs any time an existing project is passed into the component.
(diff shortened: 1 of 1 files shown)


src/components/projects/CreateProjectForm.vue
```diff
@@ -913,6 +916,10 @@ export default {
       'resetProjectForm',
       'loadProject'
     ]),
+    // Set store value for min_zenith_dist if it exists in mount config
+    load_default_zenith_from_mount () {
+      this.$store.commit('project_params/min_zenith_dist', this.selected_mount_config?.default_zenith_avoid ?? 0)
```
mgdaily, line 921: What are the `?` and `??` for?

mgdaily (approved):

PR #122, opened 2023-11-14
Update/observe page UI changes

LTDakin:
Wayne's UI request: move the telescope options to their own tab and to comment out leader follower buttons

Moved the options that form the block of controls ranging from manual telescope to pointing reference to a tab located next to controls named "Telescope Config"

In addition commented out the Leader and Follower tabs

As you can see below the config for telescope that was previously on top is now on its own tab that is only visible when you are an admin or owner

BEFORE:
Screenshot 2023-11-14 at 11 09 24 AM

AFTER:
[
Screenshot 2023-11-14 at 11 04 56 AM
](url)

### DIFF
--- a/src/components/sitepages/SiteData.vue
+++ b/src/components/sitepages/SiteData.vue
@@ -35,148 +35,11 @@
 
       <!-- Collapsible panels on the right of the image -->
       <div class="image-tools-area">
-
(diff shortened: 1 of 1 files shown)


jnation3406 (approved):
Looks good to me

PR #123, opened 2023-11-15
removed observing side replace with ascending, descending

LTDakin:
Request: Allow both ascending and descending to be selected instead of just one or the other

Previously we had a state variable called observing_side that could only be ascending OR descending. But some observations would like to follow a target as its ascending and then switch to continue following while descending. To do this we need to allow for both options to be selected independently. 

Default behavior will be true for both values

replaced the state variable observing_side from project_params, instead to support the selection of both ascending and descending the field was replaced with two checkboxes and the state was replaced with two new states 'ascending' and 'descending' that will be sent to site code instead

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -433,23 +433,12 @@
         <div class="flex-row">
           <b-field label="Observing Side">
             <b-field>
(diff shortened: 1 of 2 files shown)


mgdaily (approved):
Looks good!

PR #124, opened 2023-11-15
Feature/auto centering telescope button

LTDakin:
Michael and Wayne requested to add an OBS setting in telescope config for auto center that they will use to run some of Michael's auto centering site code. 

Updates: 
- added new CommandButtons for auto centering
- new mixins for sending the obs_configure_auto_center_on and obs_configure_auto_center_off commands to jobs api
- getter to update the status of the auto_center for the ui

need to look into why it says missing, but believe it is because the site code doesn't catch this job yet since the job is bring successfully sent to the jobs dybamodb 

Below in the image you can see the new button in the UI, and in the inspect tab you can see the new command being sent.

Screenshot 2023-11-15 at 3 24 49 PM

The record stored in the jobs dynamodb table 
Screenshot 2023-11-15 at 4 31 51 PM

### DIFF
--- a/src/components/sitepages/SiteData.vue
+++ b/src/components/sitepages/SiteData.vue
@@ -269,6 +269,32 @@
                   </CommandButton>
                 </div>
               </div>
+
+              <div class="obs-config-control-group">
+                <div class="obs-config-title">
+                  Auto Center
+                </div>
+                <StatusVal
+                  :status-item="autoCenter"
+                  style="width: 100%;"
+                />
+                <div class="obs-config-command-button-group">
+                  <CommandButton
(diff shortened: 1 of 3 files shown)


LTDakin:
Wanted to check missing message was not a frontend software responsibility, its on the api https://status.photonranch.org/status/{site}/complete_status or /{site}/obs_settings. Going to check the api to see how it receives obs_settings from site and if it can handle dynamically adding an auto_center setting or backend needs to change something.

LTDakin:
Clear on the backend, up to site to push obs settings with auto_center included

jnation3406 (approved):
looks good to me

PR #125, opened 2023-11-28
Removing intermediate smart stack check

LTDakin:
Since the backend now does not send intermediate smart stacks this code is no longer needed and can be removed.
Merging this before the paired PR isn't breaking, but then intermediate smart stacks will appear in the observe tab. 

Link to PR that adds intermediate smart stack filtering to the backend api -> https://github.com/LCOGT/photonranch-api/pull/71


mgdaily (approved):

capetillo (approved):
I am totally okay with this. Looks good to me!

PR #126, opened 2023-12-02
Feature/zoom values

capetillo:
### DESCRIPTION

**Background:**

Both Michael and Wayne have [expressed interest](https://lcogt.slack.com/archives/CLAQ9U79B/p1698092846190839?thread_ts=1698043441.211949&cid=CLAQ9U79B) in having adjusted values for `width` and `height` if certain `zoom` values were selected. There are still a few tweaks to make, like what values to have for `Big sq.` and `Small sq.`, but in the meantime, the current implementation works with the information I've been provided.

**Implementation:**

There are several steps as to how the adjusted values were implemented:

1. In `CreateProjectForm` there's a function called `getSizeDegrees`. This function performs the mathematical operations necessary to obtain the adjusted width and height values (at 100%) for some zoom selections through the api response of `site_config` (i.e. `global_config`)
2. Then, we check the limits for the zoom selections that don't have a preset value. This is done in `getMosaicLimits`. An important note is that if `getSizeDegrees` doesn't return a value because some sites don't have the configuration for it in the api, then the limit is set to `5`
3. The next step is to adjust the `width` and `height` preset values depending on the zoom selection. This is just getting the percentages of the aforementioned 100% value from `getSizeDegrees`

The next step is to append these values to the `exposures` object so that everything gets sent correctly to the backend

1. First, I added a watcher that watches for changes in  `exposures` and updates the values of `width` and `height` to the adjusted values by comparing `oldExposures` to `newExposures`
2. Then, I added an empty object called `customZoomValues` under `data`. This object gets manipulated in `updateCustomZoomValue`. The purpose of this function is to save values so that when user selects a zoom value that doesn't have an adjusted `height` and `width` and writes in a value, then selects a different zoom, then goes back to the same zoom, the value remains. 

### Visuals:

Here's a video of all the new functionality 

https://github.com/LCOGT/ptr_ui/assets/54489472/eb2374ca-eaaa-4118-bc06-ea1ab7325da2


wrosing:
This has to do with the older 36 x 36 mm 16803 chips.

W

src/components/projects/CreateProjectForm.vue
```diff
@@ -1230,7 +1397,8 @@ export default {
       'available_sites_simulated',
       'filter_wheel_options',
       'default_filter_selection',
-      'selected_mount_config'
+      'selected_mount_config',
+      'get_camera_config'
```
mgdaily, line 1401: can we call this just `camera_config`? just to match the other naming conventions for the getters

src/store/modules/site_config.js
```diff
@@ -42,6 +42,16 @@ const getters = {
     return state.global_config[state.selected_site]
   },
 
+  get_camera_config: state => {
```
mgdaily, line 45: maybe rename to camera_config?

src/components/projects/CreateProjectForm.vue
```diff
@@ -362,34 +363,54 @@
...
                 @input="val => exposures[n-1].width = val"
               />
+              <b-numberinput
+                v-else
+                :value="Number(exposures[n-1].width)"
+                :class="getSymbol(exposures[n-1].zoom)"
+                size="is-small"
+                :disabled="true"
+                type="number"
+                @input="val => exposures[n-1].width = val"
```
mgdaily, line 385: I think for the input handler, we could instead use

```
v-model="exposures[n-1].width"
```

which _should_ map exposures[n-1].width to the value of number input? 

https://vuejs.org/guide/components/v-model.html#component-v-model

src/components/projects/CreateProjectForm.vue
```diff
@@ -362,34 +363,54 @@
...
                 @input="val => exposures[n-1].height = val"
               />
+              <b-numberinput
+                v-else
+                :value="Number(exposures[n-1].height)"
+                :class="getSymbol(exposures[n-1].zoom)"
+                size="is-small"
+                :disabled="true"
+                type="number"
+                @input="val => exposures[n-1].height = val"
```
mgdaily, line 412: I think for the input handler, we could instead use

```
v-model="exposures[n-1].height"
```

which _should_ map exposures[n-1].height to the value of number input? 

https://vuejs.org/guide/components/v-model.html#component-v-model

mgdaily (commented):
Some comments - very small changes generally, but a few things I noticed

PR #127, opened 2023-12-06
Update pointing information to pull from mount status

mgdaily:
Telescope status is intended for other things, and was always a direct copy of the mount status - we had always relied on pointing being part of the telescope status, but it is no longer part of that, so switch the frontend to pull pointing information from the mount status instead.

This has been tested locally, and pointing information now shows at all sites!


LTDakin (approved):
ltgm

PR #129, opened 2023-12-06
Got rid of a repetitive function

capetillo:
`getBigSquareValues` is the exact same as `getSizeDegrees` so I got rid of it and added a check in `adjustSize`. The check is if zoom is `Full` or `Big sq.` then return the appropriate value.


mgdaily (approved):
lgtm

PR #130, opened 2023-12-07
Feature/aladin fov scaling

LTDakin:
Feature request: Aladin View should update FOV when changing sites

Changes
- moved aladin initialization into its own function for better readability. 
- set Aladin Fov inside sitecode's watcher so that when site changes FOV is re set for the new site
- added a new getter to camera config to calculate camera fov in degrees
- renamed getters to follow project conventions
- updated renamed getters in createProjectForm
- changed default view from M33 to M17 : )

https://github.com/LCOGT/ptr_ui/assets/54085254/910640d3-fd37-42d2-b260-adefbb044d0f

### DIFF
--- a/src/components/sitepages/SiteTargets.vue
+++ b/src/components/sitepages/SiteTargets.vue
@@ -681,15 +681,19 @@ export default {
     this.$loadScript('https://aladin.u-strasbg.fr/AladinLite/api/v2/latest/aladin.min.js')
       .then(async () => {
         // Default: load aladin on m33, but use coords in mount fields if possible.
-        let target = 'M33'
+        let target = 'M65'
         if (parseFloat(this.mount_ra) && parseFloat(this.mount_dec)) {
           target = `${15 * this.mount_ra} ${this.mount_dec}`
         }
 
+        // Defaults FOV of 1, use pixel size of camera if present
+        const fov = this.autoAladinFov()
+        console.log('fov: ' + fov)
+
         // Initialize Aladin
         this.aladin = A.aladin('#aladin-lite-div', {
           survey: 'P/DSS2/color',
-          fov: 1,
+          fov,
           target,
           cooFrame: 'ICRSd',
           showFullscreenControl: false,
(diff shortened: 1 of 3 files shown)


src/components/sitepages/SiteTargets.vue
```diff
@@ -748,6 +712,46 @@ export default {
...
+            showGotoControl: false,
+            showSimbadPointerControl: true
+          })
+          // cheap way to sync the skymap and mount coordinates to the aladin view.
+          setTimeout(this.set_coordinates_from_aladin, 1000)
+
+          // The following code creates an overlay element that hides aladin from mouseevents.
+          // This is mainly to let the user scroll the sidebar without aladin zooming instead.
+          // If the user clicks on aladin, then normal mouse behavior returns.
+          const scroll_hide_overlay = document.createElement('div')
```
capetillo, line 740: Here's my suggestion for how to handle this in a more maintainable and readable way:

Under data, you can return the following

```
 aladin: null,
    isOverlayVisible: true,
    overlayStyle: {
      height: '100%',
      width: '100%',
      position: 'relative',
      zIndex: 4,
```

then have a handler method, maybe called `handleOverlayClick` and here you change `isOverlayVisible` to `false`. You then can write a dynamic element in the template, something like: 

```
<div id="aladin-lite-div">
  <div 
    v-if="isOverlayVisible" 
    :style="overlayStyle" 
    @click="handleOverlayClick">
  </div>
</div>
```

I don't entirely understand the intended behavior, so adapt it as needed. But that's my main suggestion

LTDakin: I didn't really understand it either as it's Tim's code, but I can look into your suggestions if you think it needs rewriting

capetillo: why does this appear as new code?

LTDakin: when you move code in git it shows a deletion at one area and then an addition in another. So bringing this code into its own function causes the lines from 681-717 to appear as deleted and 715-754 to appear as inserted. When in fact it was separating it into its own method 'intializeAladinView()' at line 618. Green = new code is not always the case in Github.

mgdaily (approved):
Looks good - nice work @capetillo and @LTDakin for keeping the dependent branches connected up in your head! It meant we've been able to deliver features pretty quickly :)

src/components/sitepages/SiteTargets.vue
capetillo: If it was like that, then I think it's fine. I didn't know this was a github behavior. The more you know! Looks good to me, then!

capetillo (approved):

PR #132, opened 2023-12-15
Fixed adjusted width and height values

capetillo:
### Description
Background:
Wayne [reported](https://lcogt.slack.com/archives/CLAQ9U79B/p1702675691181819) that `width` and `height` for the `Full` zoom selection were not the correct values
 
**Implementation:**
I wrote two simple functions. One that gets full width and one that gets full height. This is done by getting either `camera_size_x` or `camera_size_y`, multiplying by `pixel_scale` and dividing by 3600. If any of these values doesn't exist, we return `1`. I also wrote a function to get `Big sq.` values since it's not the same value as `Full`. Pretty simple fix :-)

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -874,8 +874,8 @@ export default {
     this.load_default_zenith_from_mount()
 
     // initializing width and height to the appropriate value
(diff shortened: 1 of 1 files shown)


mgdaily (approved):
Nice quick fix! Thanks for explaining it all to me

PR #134, opened 2023-12-19
Feature/Zoom Dropdown under Camera Tab

capetillo:
### Description:

**Background:**
Wayne made a request [here](https://lcogt.slack.com/archives/CLAQ9U79B/p1702935461713609) and provided details about his request [here](https://lcogt.slack.com/archives/CLAQ9U79B/p1702950867871319). Basically, he requested for a `zoom` dropdown menu under the Camera tab so that users can have a similar experience of being at a telescope. The `zoom` should change the eyepiece and achieve the difficult to replicate experience that cheap telescopes claim to be able to provide.

**Implementation:**
The implementation was fairly simple. I added a dropdown menu to the `Camera` component that loops through the `zoom_options`. These options are in an array under `data` and they are `Full`, `Small sq.`, `71%`, `50%`, `35%`, `25%`, `18%`, `12.5%`, `9%`, and `6%`.  Then in this same component under computed properties I added a new one called `zoom_options_selection`. This first gets the getter `zoom_options_selection` from the file `commands_params` and then updates the value of `zoom_options_selection` in the Vuex store when the user changes the `zoom` value in the UI. 
The next step was to add the getters to `command_params` so that the aforementioned values can be set and mutated. This is also where I set the default value of zoom selections to `Full`. 

### Visuals:

**Default zoom value in the UI and the store**
Screenshot 2023-12-19 at 3 25 38 PM

**Recording of zoom values updating in the store as user selects different values**

https://github.com/LCOGT/ptr_ui/assets/54489472/a4d29cf7-305a-4038-bb4d-4f8aa2f92cda

### DIFF
--- a/src/components/InstrumentControls/Camera.vue
+++ b/src/components/InstrumentControls/Camera.vue
@@ -130,6 +130,26 @@
     </b-field>
 
     <b-field
+      horizontal
+      label="Zoom"
+    >
+      <b-select
+        v-model="zoom_options_selection"
+        placeholder="select zoom..."
+        size="is-small"
+      >
+        <option
+          v-for="(zoom, index) in zoom_options"
+          :key="index"
+          :value="zoom"
+          :selected="index === 0"
+        >
+          {{ zoom }}
+        </option>
+      </b-select>
+    </b-field>
+
+    <b-field
       v-if="camera_can_bin"
       horizontal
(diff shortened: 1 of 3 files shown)


LTDakin (approved):
Good fix and looks good to merge

PR #139, opened 2024-01-09
Draw sublunar point on ptr world map

LTDakin:
Draws the moon to the world map on the ptr home screen 

using an good enough equation supplied by Tim Lister -> https://stjarnhimlen.se/comp/tutorial.html#7. 
Tested against this website to check accuracy -> https://www.timeanddate.com/worldclock/sunearth.php

Hard to test since I don't understand the math too much, so will keep an eye on it for a week to make sure it stays accurate and no errors come up.

Screenshot 2024-01-09 at 1 53 56 PM


capetillo (approved):
Great work! Big math behind it. Good for you for crushing it!

capetillo (commented):
I had approved before, but  the moon isn't moving.

src/components/maps/nite-overlay.js
```diff
@@ -114,6 +120,72 @@ const nite = {
...
+    // Initial orbital elements
+    const N = rev(125.1228 - 0.0529538083 * d);
+    const i = 5.1454;
+    const w = rev(318.0634 + 0.1643573223 * d);
+    const a = 60.2666;
+    const e = 0.054900;
+    const M = rev(115.3654 + 13.0649929509 * d);
+
+    // Normalize M
+    const normalizedM = rev(M + 129 * 360);
```
markBowman, line 154: I'm not sure this line is necessary.

src/components/maps/nite-overlay.js
```diff
@@ -114,6 +120,72 @@ const nite = {
...
+
+    // Compute eccentric anomaly (E)
+    let E0 = normalizedM + (180 / Math.PI) * e * Math.sin(normalizedM) * (1 + e * Math.cos(normalizedM));
```
markBowman, line 157: I thought javascript Math.sin() etc used radians rather than degrees

src/components/maps/nite-overlay.js
```diff
@@ -114,6 +120,72 @@ const nite = {
...
+    let E1;
+
+    do {
+      E1 = E0 - (E0 - (180 / Math.PI) * e * Math.sin(E0) - normalizedM) / (1 - e * Math.cos(E0));
+      E0 = E1;
+    } while (Math.abs(E0 - E1) > 0.005);
```
markBowman, line 163: Is this guaranteed to converge? Do we need to limit the number of iterations?

src/components/maps/nite-overlay.js
```diff
@@ -114,6 +120,72 @@ const nite = {
...
+
+    const E = E1;
+
+    // Rectangular coordinates in the plane of the lunar orbit
+    const x = a * (Math.cos(degToRad * E) - e);
```
markBowman, line 168: Yeah, this looks right - converting degrees to radians before taking the cosine

src/components/maps/nite-overlay.js
LTDakin: Looking at that now, yeah that doesn't make any sense. Theres a line in the article that says 'To normalize M we had to add 129*360 = 46440 degrees.' But if it outputs the same value then I will remove it

src/components/maps/nite-overlay.js
LTDakin: Not sure, but should limit it to be safe

src/components/maps/nite-overlay.js
```diff
@@ -114,6 +120,72 @@ const nite = {
...
+    const zeclip = r * Math.sin(degToRad * (v + w)) * Math.sin(degToRad * i);
+
+    // Convert to ecliptic longitude, latitude, and distance
+    const eclipticLongitude = Math.atan2(yeclip, xeclip) * radToDeg;
+    const eclipticLatitude = Math.atan2(zeclip, Math.sqrt(xeclip * xeclip + yeclip * yeclip)) * radToDeg;
+
+    console.log(eclipticLongitude)
+    console.log(eclipticLatitude)
+
+    return new google.maps.LatLng(eclipticLatitude, eclipticLongitude)
```
markBowman, line 187: Wow! That's hard math! Well done.

src/components/maps/nite-overlay.js
markBowman: I think they were just giving an example of what the normalization does, rather than adding an extra step

src/components/maps/nite-overlay.js
LTDakin: Can't say I understand it at all as I copied from the textbook Tim gave me, but thanks!

LTDakin:
Ah I see why it's not working. This formula gives ecliptic longitude and latitude when I need to get equatorial. Will look for a formula for that or ask Tim again.

LTDakin:
Closing since its been a year and this feature isn't requested anymore

PR #140, opened 2024-01-12
Feature/add estimate focus offset option

LTDakin:
Request: https://lcogt.slack.com/archives/CLAQ9U79B/p1705011218124859

mimicked a lot of the code from auto focus. Create a new script and component for displaying the script. Added the script name to the state of sequence scripts in the vuex store. tested that it would send a site request with the script name.

Eventually once Micheal implements the scripts or confirms it has been implemented you should go back and remove it from the 'notImplementedScripts' array in 'Sequencer.vue'

Screenshot 2024-01-12 at 3 09 07 PM


capetillo (approved):
Because I don't really know what this does or the purpose of it, I'm comparing your new code to `focusFine` and it seems like the logic is pretty much identical. LGTM

PR #142, opened 2024-02-02
Validation fixes for Create Project

LTDakin:
Micheal mentioned that projects could be created without a target name. This PR fixes that as well as a lot of other bugs I found along the way which include: hanging on empty search, no error messages for missing ra and dec, and not verifying fields when modifying a project

Changes
- new field 'Name' splitting up the functionality of the Object Search field.
- check for target name in verifyForm() so that a target name is required to create or modify a project
- warning messages for missing ra and dec fields
- error message for unknown targets are now info instead of an error

Fixes
- fixed naming for projectRa and projectDec to targetRa and targetDec
- added a VerifyForm call to modifyProject, previously it was missing
- removed unnecessary b-field tag
- removed redundant logic for checking ra and dec fields are empty in verifyForm()
- combined two separate checks for warnings into one if/else
- check for if result.Target exists to prevent hanging on an blank target search

Screenshot 2024-02-06 at 12 04 29 PM

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -174,39 +174,52 @@
           <b-field
             label="Name"
             style="width: 230px;"
+            :type="{'is-warning': warn.target_name}"
           >
-            <b-field>
-              <b-input
-                v-model="targets[0].name"
-                :disabled="!targets[0].active"
-                style="max-width: 130px;"
-                class="project-input"
-              />
-              <p class="control">
-                <b-button
-                  class="button"
-                  :loading="object_name_search_in_progress"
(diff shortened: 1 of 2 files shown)


src/components/projects/CreateProjectForm.vue
```diff
@@ -1008,6 +1021,7 @@ export default {
 
     verifyForm () {
       if (this.project_name === '') { this.warn.project_name = true }
+      if (this.targets[0].name === '') { this.warn.target_name = true }
```
capetillo, line 1024: Out of curiosity, what happens if the user enters multiple spaces?

capetillo (commented):
one small comment : )

mgdaily (approved):

PR #143, opened 2024-02-05
Small Naming changes for target page and search box

LTDakin:
Wayne wanted the Targets page to be named to Sky Map and Object name in camera tab to just be object to match the search fields in the sky map tag and the projects tab.

Screenshot 2024-02-05 at 10 52 41 AM


capetillo (approved):
Looks good!

PR #144, opened 2024-02-15
Add Read-Only Project View, Add Username to Projects Table

mgdaily:
This PR adds an "inspect" button to the projects table, which opens up a read-only view of the CreateProjectForm, pre-populated with the data that needs to be displayed.

_RightAscensionInput/DeclinationInput/ProjectSitesSelector:_
* Add "disabled" prop so the underlying input fields may be disabled

_CreateProjectForm_
* Go through and conditionally disable each input/field/checkbox if the form is read-only
* Adjust the props to have defaults so we can set read-only to false by default

_UserProjectsTable_
* Add "inspect" button to table row, which fetches the project, saves whatever is currently entered into the form into the store to be reloaded after the inspector is closed, and emits an `inspect-project` signal along with the project data for the so the parent SiteProjects component can pop up and render the modal.
* Add username as column, make sortable

_SiteProjects_
* Add modal that is hidden by default, but will be shown and render a read-only CreateProjectsForm, with all data from the project that has been selected for inspection.

_project_params store module_
* Add a `draft_project_params` store variable to hold the draft of the project that's been entered into the project form prior to the user clicking "inspect," so that they may be re-loaded when the user closes the inspection tab. Unfortunately the CreateProjectForm is tightly coupled to the vuex store and whenever we render a new instance of that component, it overwrites the data in the project_params module of the store - so we need to keep a history of anything that was entered into the form, if the user had done that.
* Add actions to both save and re-load a project draft from the store.

**Video!**

https://github.com/LCOGT/ptr_ui/assets/7182533/43fde863-91c5-4560-8500-927f069266da

### DIFF
--- a/src/components/FormElements/DeclinationInput.vue
+++ b/src/components/FormElements/DeclinationInput.vue
@@ -9,10 +9,12 @@
         v-model="localDec"
         :size="size"
         lazy
+        :disabled="disabled"
       />
       <b-select
         v-model="units"
         :size="size"
+        :disabled="disabled"
       >
         <option value="deg">
           deg
@@ -27,10 +29,10 @@
 
 <script>
 export default {
-  props: ['value', 'size'],
+  props: ['value', 'size', 'disabled'],
   data () {
     return {
-      localDec: null,
+      localDec: this.value,
       units: 'deg',
       hasError: false,
       errorMessage: ''
--- a/src/components/projects/UserProjectsTable.vue
+++ b/src/components/projects/UserProjectsTable.vue
@@ -22,7 +22,6 @@
       default-sort-direction="desc"
       detailed
       class="my-table no-margin"
-      @click="inspectProject($event.row.project_name, $qevent.row.created_at)"
     >
       <b-table-column
         v-slot="props"
@@ -30,6 +29,15 @@
         label="project name"
       >
         {{ props.row.project_name }}
+      </b-table-column>
+
+      <b-table-column
+        v-slot="props"
+        field="project_creator.username"
+        label="user"
+        sortable
+      >
+        {{ props.row.project_creator ? props.row.project_creator.username : '' }}
       </b-table-column>
 
       <b-table-column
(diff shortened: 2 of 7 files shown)


src/store/modules/project_params.js
```diff
@@ -172,7 +175,14 @@ const mutations = {
   start_date (state, val) { state.start_date = val },
   long_stack (state, val) { state.long_stack = val },
   smart_stack (state, val) { state.smart_stack = val },
-  defocus (state, val) { state.defocus = val }
+  defocus (state, val) { state.defocus = val },
+  draft_project_params (state, val) { state.draft_project_params = val }, // project params that have saved off in saveProjectDraft
+  rewrite_state (state, val) {
+    // loop through and rewrite the state
+    for (const stateItem in val) {
+      Vue.set(state, stateItem, val[stateItem])
```
mgdaily, line 183: In order to restore the state from the draft_project_params, we need to loop through every attribute in the draft params and set the corresponding state variable.

src/components/projects/UserProjectsTable.vue
```diff
@@ -209,6 +224,22 @@ export default {
...
+      const request_params = {
+        project_name,
+        created_at
+      }
+      const project_endpoint = this.$store.state.api_endpoints.projects_endpoint + '/get-project'
+      axios.post(project_endpoint, request_params).then(response => {
+        const project_loader = {
+          project: response.data
+        }
+        this.$store.dispatch('project_params/saveProjectDraft')
```
mgdaily, line 237: When the new CreateProjectForm is rendered with project data, it re-writes the current project data in the store, so we need to first save off anything that has been entered before we destroy it, so we can load it back later!

src/components/FormElements/DeclinationInput.vue
```diff
@@ -27,10 +29,10 @@
 
 <script>
 export default {
-  props: ['value', 'size'],
+  props: ['value', 'size', 'disabled'],
   data () {
     return {
-      localDec: null,
+      localDec: this.value,
```
mgdaily, line 35: Here we default the localDec value to the prop that was passed in - I was having an issue when I was re-rendering the component, the dec wasn't displaying even though the value was being passed. So set the default to the value passed in.

LTDakin (approved):
Like the reuse of the form component, and how closing the inspection reloads from drafts. Lgtm

capetillo (approved):
beautiful! also thank you for the video, that's very helpful : )

PR #145, opened 2024-02-21
Re-order buttons on project row. Add padding to delete button.

mgdaily:
This is a simple change based on Wayne's message.

> Great. Please put inspect top left, Clone next to it, modify below and then keep delete the same size but move it over to right so its right edge coincides with the right edge of the Clone button.
That will minimize accidental hits of the delete button.

Screenshot:
image


capetillo (approved):
margin-left is pure genius. Looks great!

PR #147, opened 2024-03-18
Feature/display skymap values

capetillo:
## FEATURE: Display `transition time` and `airmass at transition` values

**Background:**
Wayne [made a request here](https://lcogt.slack.com/archives/CLAQ9U79B/p1709440661804119) to display a few values. `airmass` already existed so that was not necessary to implement. I chatted with him and he told me to add fields and that he would take care of getting the values. In the meantime, he wants 'NaN' displayed. 

**Implementation:**
Followed the logic of what the rest of the fields in Site status footer were doing. What I'm not sure of, is if the value itself will be a number and if it makes sense to parse it and fix it. But in the meantime, this can remain as is and it returns 'NaN'. 

### VISUALS
**On the right, you can see the 2 new values `trans. time` and `trans. airmass`**
Screenshot 2024-03-18 at 11 04 51 AM


mgdaily (approved):
clean!

PR #148, opened 2024-03-19
New script setting equatorial sweep with two numerical input fields

LTDakin:
Per Wayne's request, new script field for Equatorial Sweeps. Min 0 for both, max none, default to 12 and 30.

Screenshot 2024-03-19 at 11 57 04 AM


capetillo (commented):
Logic looks good. My one comment is that I think the arrows should look more like the screenshot and the text should be to the left of it rather than above to maintain structure: 
Screenshot 2024-03-19 at 2 40 55 PM

mgdaily (approved):

PR #150, opened 2024-03-25
Fix/min exposure

capetillo:
## QUICK FIX: Minimum exposure

Michael discovered that users cannot input decimals for exposures. This was quickly changed by changing the `min` value of the Exposure field to be `0.000001` and having a step of `any`. This step defaults to 1 but if the user wants to manually input something else, then that can be done too.

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -322,7 +322,7 @@
                 size="is-small"
                 :disabled="!exposures[n-1].active || read_only"
                 type="number"
-                min="0"
+                min="0.000001"
                 max="100000"
                 step="any"
               />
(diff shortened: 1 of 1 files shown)


src/components/projects/CreateProjectForm.vue
```diff
@@ -322,12 +322,10 @@
                 size="is-small"
                 :disabled="!exposures[n-1].active || read_only"
                 type="number"
-                min="0"
+                min="0.000001"
```
mgdaily, line 325: This is fine. Honestly in practice a min of 0.1 with step of 0.1 would probably suffice. I don't know that any of these cameras can control the exposure time in microseconds lol

mgdaily (approved):

src/components/projects/CreateProjectForm.vue
mgdaily: But doing what Michael said is totally fine, just go with that.

PR #151, opened 2024-04-01
Do not render email addresses for anonymous users

mgdaily:
This is based off a request from Michael to not show email addresses to non-admin users.

The only users that should be able to see any sort of email address are the users that own the calendar block or an admin.

User as admin:

image
Screenshot 2024-04-01 at 11 12 30 AM

Anonymous user:

image
image


capetillo (approved):
Looks good to me! Thank you for taking care of this one :-)

PR #152, opened 2024-04-11
fix to overflow on exposure panel in create project form

LTDakin:
Fixed some overflow by changing display to flex and removing code that was trying to insert a white space label to fix the spacing of the check-boxes. Instead added a CSS selector for the first checkbox and shifted it down. 

**Before**
Screenshot 2024-04-10 at 3 59 00 PM
**After**
Screenshot 2024-04-11 at 12 54 39 PM


LTDakin:
in the spirit of Mark's talk I'm just going to merge this

PR #154, opened 2024-04-15
Replaced longstack with substack

LTDakin:
Replaced long stacks with sub stacks at Wayne and Michaels request. Tested it to make sure that its being sent up as a key in the create project json. Can check the tst site and view my project that is testing substacking.


capetillo (approved):
Nice! LGTM 👍

PR #155, opened 2024-04-29
set sub_stack to be default to on since its prefered

LTDakin:
changes Sub stack to be defaulted to 'on' in the camera and project settings. Changed command params which is the default state for fields and changed the value in resetProjectForm so that when the project form is reset it will be true.


capetillo (approved):
LGTM! Bam

PR #156, opened 2024-05-28
Fix/project values

capetillo:
## FIX: >Project values

**Background:**
Wayne[ made a request](https://lcogt.slack.com/archives/CLAQ9U79B/p1716659993681929) to switch the values for `width` and `height` under the `Projects` tab and allow for up to 2 decimal places for `airmass` under `Advanced Options`.

**Implementation:**
For the first part of the request, I just changed `camera_size_x` for `camera_size_y` when we get `width` and the opposite for when we get `height`. 
Then for the second part of the request, I added a method that rounds max airmass called `roundMaxAirmass` to the nearest 2nd decimal and it's invoked in the `blur` event of the `Max Airmass` field. _**That being said**_, there is a `warn.max_airmass` that is immediately triggered but it goes away if the user clicks on the field again. Not sure if we should get rid of the warning altogether but it works without if the user enters up to 2 decimal places.

### VISUALS
**Screenshot of width now being the greater value**
Screenshot 2024-05-28 at 2 13 17 PM

**Screen recording of user entering more than 2 decimal places, warning appears, then it goes away again**

https://github.com/LCOGT/ptr_ui/assets/54489472/d8f7a4a7-c229-4cc4-8e4b-897dde75d9a3

### DIFF
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -659,7 +659,7 @@
               v-model="max_airmass"
               class="project-input"
               type="number"
-              min="0"
+              min="1"
               max="5"
               step="0.01"
               :disabled="read_only"
--- a/src/components/projects/CreateProjectForm.vue
+++ b/src/components/projects/CreateProjectForm.vue
@@ -659,9 +659,11 @@
               v-model="max_airmass"
               class="project-input"
               type="number"
-              min="1"
+              min="0"
               max="5"
+              step="0.01"
               :disabled="read_only"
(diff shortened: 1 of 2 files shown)


mgdaily (approved):
Looks good - thanks for taking care of it!

Re: the warning in your video - it looks like that's triggered because the value the user entered had more decimal places than what the step specifies, but it's validated before the rounding happens, so it displays the error before it gets a chance to round the number? Based on Wayne's message I think we can make the step quite small (0.0001) to make the arrows work, but it'll always get rounded once the blur event occurs?

PR #158, opened 2024-06-28
fixed duplicate moons

capetillo:
## FIX: Duplicate moons 

**Background:**
Wayne brought up the issue where the lunar bands weren't showing up (fixed now) and the moon icons were duplicated.

**Implementation:**
In the function `dayRender`, I've added a query selector that finds the header cell (i.e. where we actually want the moon icons to render). Then if that exists, I create `moonIconDataAttr` which is unique to each date. Then I look for other existing moon icons and remove them. Lastly, I create the moon icon and append it to the header cell. 

### VISUALS
**Before: duplicate moons**
Screenshot 2024-06-28 at 2 17 05 PM

**After: one moon per column**
Screenshot 2024-06-28 at 2 16 34 PM

### DIFF
--- a/src/components/calendar/TheCalendar.vue
+++ b/src/components/calendar/TheCalendar.vue
@@ -669,6 +669,9 @@ export default {
     },
 
     dayRender (dayRenderInfo) {
+      if (this.fullCalendarApi?.view?.type === 'timeGridWeek') {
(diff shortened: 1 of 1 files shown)


mgdaily (approved):
yay!

PR #160, opened 2024-07-02
Fix/aladin view

capetillo:
## FIX AND REARRANGEMENTS: Aladin view

**Background:**
Wayne requested to change a few aspects of the Aladin section under the `Sky Map` tab. His requests included: 1. removing search for object under Aladin 2. removing the `slew to RA/DEC` and `slew & center` buttons 3. rearrange the order of the boxes in the accordion 4. when searching for an object, automatically populate the `Hour Angle`, `Azimuth`, and `Altitude`

**Implementation:**
I got rid of the unwanted components and rearranged as needed. For the last request this is what I did: 
In the `helpers.js` file I corrected `eq2altaz` by converting `ha` (hour angle) to degrees. With this function, we can get `azimuth` and `altitude`.
Then, I added `eq2hourangle` function to return hour angle in hours.
Then these two helper functions are used in `handle_coordinate_search_results` in the `Telescope` component.

### VISUALS
**Before-- you can see that after searching for an object, hour angle, azimuth, and altitude remain blank**
Screenshot 2024-07-02 at 3 10 50 PM

**After-- fields are automatically filled out :)**
Screenshot 2024-07-02 at 3 12 35 PM


mgdaily (approved):
Thanks so much!!!

PR #162, opened 2024-09-17
UI Requests

timbeccue:
Adding a few modifications to the UI from Wayne's requests:

- Add some additional image download options, but they should all still just download the same jpg. Wayne wanted to be able to simulate what might be available in the future while he thinks through how to support them. 
Screenshot 2024-09-17 at 3 20 51 PM

- On the telescope control tab, move the "Slew to..." button above the status display (below the Alt/az slew inputs), and add a park button in front. 
Screenshot 2024-09-17 at 3 21 29 PM

- In the form to create projects, add a field to the exposures form called "Repeat" that accepts integers in range [0,150] and with default of 0. Should be between the 'exposure' and 'filter' inputs. 
Screenshot 2024-09-17 at 3 22 31 PM


LTDakin (approved):
looks good!

PR #164, opened 2024-10-08
Add 'cam_note' and 'tel_note' to command payloads

timbeccue:
The cam_note already existed under a different name, so I renamed those occurrences.
I added tel_note to the telescope command tab. 

- cam_note and tel_note are both keys in the optional_params dict that is part of the command payload
- they are not included if the value is an empty string
- tel_note is only considered if the mount command is the ‘go’ command (all green circled buttons in the attached screenshot). tel_note is never included for the other mount commands (red circled buttons in screenshot) like park, slew to flat position, etc.
Screenshot 2024-10-08 at 7 06 30 PM

### DIFF
--- a/src/components/InstrumentControls/Camera.vue
+++ b/src/components/InstrumentControls/Camera.vue
@@ -273,10 +273,10 @@
 
     <b-field
       horizontal
-      label="Note"
+      label="Cam Note"
     >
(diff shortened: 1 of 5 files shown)


wrosing:
All good, thanks.

PR #165, opened 2024-10-09
Add tick marks and a centered circle to image crosshairs

timbeccue:
Added a few enhancements to the image crosshairs. 

Each crosshair has been given tick marks, which are spaced so that 6 ticks fit on each half of the horizontal axis (13 total if you count the center). The spacing for the ticks on the vertical crosshair are the same as those on the horizontal one. 

There's also a circle that is centered on the image with radius of 20% of the width. 

Screenshot 2024-10-08 at 8 44 16 PM


wrosing:
Thanks that will really help!

mgdaily (approved):

